Issue #11, opened 2019-11-13
Update overcooked_js to be compatible with order_list of None


micahcarroll:
This is necessary for example when exporting trajectories from python to visualize them in overcooked-demo.

Issue #9, opened 2019-11-13
When saving json trajectories, ensure layout naming is changed to JS format


micahcarroll:
This is necessary for example when exporting trajectories from python to visualize them in overcooked-demo.

Issue #8, opened 2019-11-12
Saving trajectories in json format fails due to arrays and int64s in returned trajectories


micahcarroll:
Make sure to format everything to lists and ints when creating trajectories. Could add type assertions for trajectory formatting check.

PR #12, opened 2019-11-19
Adding action_infos to trajectories and bug fixes

micahcarroll:
The biggest change in the python code is that now all agents return a tuple (`action`, `action_infos`) rather than just returning `action`. `action_infos` is a dictionary that can contain additional information (such as the action probs). This dictionary will then be added to returned trajectories' standard `ep_infos` key.

Also introduced a `metadata` field for standard trajectories, that one can compute and store arbitrary functions of trajectory data.

Both of these changes are to make the analysis of trajectory data easier to do, no matter what one is specifically trying to analyze.

Bug fixes:
- Standardized naming of layouts across python and javascript (now equal to paper), fixing issue #9.
- Fixes bug in saving of trajectories in the python side #8.
- Now Javascript code works also when not passed any order list (assumes infinitely many "any" orders) #11. 
- Also fixed a visualization bug due to which pots were not being displayed as "cooked", although they were in a cooked status, that also affected Overcooked-demo.


overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -59,6 +79,7 @@ def mdp_generator_fn():
                 display=display
             )
         
+        # variable_mdp = (mdp_choices is None) or (mdp_choices is not None and len(mdp_choices) != 1)
```
rohinmshah: Delete this since the computation is now happening in `overcooked_env.py`

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -238,26 +252,27 @@ class Overcooked(gym.Env):
...
         """
         if baselines:
             # NOTE: To prevent the randomness of choosing agent indexes
             # from leaking when using subprocess-vec-env in baselines (which 
             # seeding does not) reach, we set the same seed internally to all
             # environments. The effect is negligible, as all other randomness
             # is controlled by the actual run seeds
-            np.random.seed(0)
+            # np.random.seed(0)
+            pass
```
rohinmshah: ... Should this entire if block and comment be deleted?

overcooked_ai_js/js/mdp.es6
```diff
@@ -368,7 +357,7 @@ export class OvercookedGridworld {
     }
 
     is_terminal ({state}) {
-        return (state.order_list.length === 0) || (state.pot_explosion)
+        return (state.order_list === null) || (state.order_list.length === 0)
```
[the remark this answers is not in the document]
micahcarroll: Yes you're totally right, thanks for catching that!

overcooked_ai_py/mdp/overcooked_env.py
micahcarroll: I feel like it was non-obvious (at least to me) that if using parallel threads it might be harder to propagate random seeds to the threads, but I guess you're right in that this is possibly baselines specific and shouldn't be in this repo.

rohinmshah: Oh I'm just saying that you currently have `if <condition>: pass` which is odd. I don't really understand what the problem is, why this used to be there, why it isn't there now, how other people are supposed to interact with this code; I'm just flagging it as a thing that's confusing me.

micahcarroll:
I've made another couple of changes (and updated the PR description) if you want to quickly take a look @rohinmshah

PR #16, opened 2019-12-09
pk-dev3 into overcooked_ai_improvements

paulk444:
Hey Micah. There a few things to flag up with this PR:

1) In my branch, in overcooked_env.get_rollouts, I had "ep_returns_sparse" for sparse returns and "ep_returns" for shaped returns (see [here](https://github.com/HumanCompatibleAI/overcooked_ai/blob/26adb30944ffaf3bbee155063278bd5a6b6d0f96/overcooked_ai_py/mdp/overcooked_env.py#L167)). But in your branch these had become "ep_returns" for sparse, and I think there was no shaped return here. I require shaped returns so I added "ep_returns_shaped". And for consistency I left the sparse as "ep_returns".

I changed “ep_returns” → “ep_returns_shaped” and “ep_returns_sparse” → “ep_returns” throughout my code. But I'm not certain that I've done this everywhere, so it's something to be aware of (mainly for me) in the future. If the rewards look incorrect then this might be why!

2) Your GreedyHumanModel returns action probs, whereas my ToMModel, and the model used by PPO during training, don't return action probs. This meant that [this line](https://github.com/HumanCompatibleAI/overcooked_ai/blob/2cdf8fe72bec460e9a0cbe51b5d165820ed93587/overcooked_ai_py/mdp/overcooked_env.py#L149) was causing me trouble. My solution was to add a parameter to all agents called self.return_action_probs, which if set to False means the actions only return the action itself, not the action probs.

3) The layout names have changed from the old names "simple", "random0" etc to the proper names in the paper. I tried to just update the names throughout my code but this proved more difficult than expected. E.g. the BC data itself uses the name "simple" rather than "cramped_room", which didn't seem easy to change. So as a temporary solution I've just made two versions of each layout, one with each name.

Otherwise I *think* my changes are uncontroversial or don't affect your code!


overcooked_ai_py/agents/agent.py
```diff
@@ -111,32 +116,32 @@ def __init__(self, agent):
...
         stochastic (Bool): Whether the agent should sample from policy or take argmax
-        action_probs (Bool): Whether agent should return action probabilities or a sampled action
+        return_action_probs (Bool): Whether agent should return action probabilities or a sampled action
         """
         self.state_policy = state_policy
         self.direct_policy = direct_policy
         self.history = []
         self.stochastic = stochastic
-        self.action_probs = action_probs
+        self.return_action_probs = return_action_probs
```
micahcarroll: I should have done this when cleaning but missed it: remove action_probs entirely from this class, and always pass in action_probs=True when calling self.state_policy.

overcooked_ai_py/agents/agent.py
```diff
@@ -205,25 +209,29 @@ class FixedPlanAgent(Agent):
     # NOTE: Assumes that calls to action are sequential (agent has history)
     """
 
-    def __init__(self, plan):
+    def __init__(self, plan, return_action_probs=True):
```
micahcarroll: Remove action_probs attribute

overcooked_ai_py/agents/agent.py
```diff
@@ -205,25 +209,29 @@ class FixedPlanAgent(Agent):
...
     
     def action(self, state):
         if self.i >= len(self.plan):
             return Action.STAY, {}
         curr_action = self.plan[self.i]
         self.i += 1
-        return curr_action, {}
-    
+        if self.return_action_probs:
+            return curr_action, {}
```
micahcarroll: Always return empty dictionary, infrastructure should always expect an (action, info_dictionary) pair as return to the action method, even if such dictionary is empty.

overcooked_ai_py/agents/agent.py
```diff
@@ -301,28 +309,31 @@ def action(self, state):
...
+        if self.logging_level >= 1:
             print("expected joint action", first_joint_action)
         action = first_joint_action[self.agent_index]
         return action, {}
 
 
 class GreedyHumanModel(Agent):
     """
+    This is Micah's GreedyHumanModel, which is slightly different to Paul's (called _pk)
+
```
micahcarroll: We should probably remove comments referring to our names specifically (as we expect that people that don't know us will be trying to use this code), and possibly find better naming conventions instead that capture the essence of the difference between the model types.

overcooked_ai_py/agents/agent.py
```diff
@@ -333,6 +344,9 @@ def __init__(self, mlp, hl_boltzmann_rational=False, ll_boltzmann_rational=False
         self.auto_unstuck = auto_unstuck
         self.reset()
 
+        # Set to true to return action probs:
+        self.return_action_probs = True
```
micahcarroll: Remove

overcooked_ai_py/agents/agent.py
```diff
@@ -373,18 +387,22 @@ def action(self, state):
...
                 action_probs = self.a_probs_from_action(chosen_action)
 
             # NOTE: Assumes that calls to the action method are sequential
             self.prev_state = state
-        return chosen_action, {"action_probs": action_probs}
+
+        if self.return_action_probs:
+            return chosen_action, {"action_probs": action_probs}
+        else:
+            return chosen_action
```
micahcarroll: Fix

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
 
         return motion_goals
 
+#
+# class AgentPair(object):
+#     """
+#     The AgentPair object is to be used in the context of a OvercookedEnv, in which
+#     it can be queried to obtain actions for both the agents.
+#     """
```
micahcarroll: Remove these commented out lines

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
...
+#             return (action_0, action_1)
+#         else:
+#             return (self.a0.action(state), self.a1.action(state))
+#
+#     def reset(self):
+#         for a in self.agents:
+#             a.reset()
+
+
+# ============================ Agents make by pk ===================================#
```
micahcarroll: Remove this

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
...
+
+class GreedyHumanModel_pk(Agent):
+    """
+    This is Paul's GreedyHumanModel, which is slightly different Micah's (just called GreedyHumanModel)
```
micahcarroll: Remove this line and rename class to something like  GreedyHumanModelImproved (choose whatever you think is most sensible)

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
...
+    Enhancements added to v2:
+    1) Drop onion if a dish is needed instead
+    2) Drop dish if an onion is needed instead
+    3) If soup on the counter then pick it up
+    3) Added parameter "perseverance", then added mechanism for resolving crashes: each agent steps out of the way with
+    a certain probability, which depends on their perseverance and how many timesteps they've been stuck.
+
+    """
+
+    # TODO: Remove any mention of tomato!
```
micahcarroll: Remove this line, and logging.info calls

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
...
+            if (not other_has_onion) and (
+                    others_min_cost >= own_min_cost):  # TODO: make it random if others_min_cost = own_min_cost ?
+                logging.info('Getting an onion: assume in better position that other player')
+                motion_goals = am.pickup_onion_actions(state, counter_objects, self.only_take_dispenser_onions)
+                motion_goals = self.sometimes_make_wrong_decision(motion_goals, state, counter_objects, am,
+                                                                  current_goal='onion')
+
+            elif other_has_onion or ((not other_has_onion) and (others_min_cost < own_min_cost)):
+                # Now find the *next* task that needs to be done, and do that
+
```
micahcarroll: Typo: "deprecated" code

micahcarroll: We shouldn't commit much deprecated & commented out code to overcooked_ai. This is another good reason to move most of your agents (and all this other code) to your project repo, so you can keep it a bit less organized till release.

overcooked_ai_py/agents/agent.py
```diff
@@ -515,3 +532,2155 @@ def ml_action(self, state):
...
+        # distributed random variable is 1/p, so set p to e.g. 0.25?
+        self.path_teamwork = path_teamwork  # Prob of considering the other agent's location in the path choice
+        self.rationality_coefficient = rationality_coefficient  # Setting to 0 means random actions; inf means always takes
+        # lowest cost path. In practice inf ~ 100
+        self.prob_pausing = prob_pausing # Probability of pausing on a given timestep, instead of acting. From a
+        # quick initial look at the human data, the humans pause very approx 50% of the time
+
+        self.return_action_probs = False  # Don't return action probs
+
+    def reset_agent(self):
```
micahcarroll: Should be simply called `reset` to conform to the Agent API as defined at the top of the file

overcooked_ai_py/agents/benchmarking.py
```diff
@@ -16,6 +16,7 @@
     "ep_dones",
     "ep_infos",
     "ep_returns",
+    "ep_returns_shaped",
```
micahcarroll: Revert this back

overcooked_ai_py/agents/view_agent.py
```diff
@@ -0,0 +1,357 @@
...
+    elif layout == 'ran1':
+        start_state = OvercookedState([P((1, 2), n), P((1, 1), n)], {}, order_list=start_order_list)
+    elif layout == 'ran0':
+        start_state = OvercookedState([P((1, 2), n), P((3, 2), n)], {}, order_list=start_order_list)
+    elif layout == 'ran3':
+        start_state = OvercookedState([P((1, 2), n), P((6, 2), n)], {}, order_list=start_order_list)
+    else:
+        start_state = OvercookedState([P((2, 2), n), P((2, 1), n)], {}, order_list=start_order_list)
+
+    if layout == 'sc1':
```
micahcarroll: If you want to commit this to overcooked_ai, it would be nice to remove all the repeats in the code first (have a way to abstract this rather than having 5 if statements with a lot of repeated code). Alternatively, you should probably use your project repo: I'll be less rigid in enforcing nice coding practice there, as it will mostly just be affecting you. This file is a great idea though! 😄

overcooked_ai_py/agents/zeroth_order_opt_active_states.py
```diff
@@ -0,0 +1,579 @@
+import time
```
micahcarroll: Again, I would move this file to your project repo, as it is currently quite specific to your project. For similar reasons to the agent.py file review (easier diff-viewing), I will do a more thorough review once it's in the PR on that repo.

overcooked_ai_py/data/layouts/random0.layout
```diff
@@ -0,0 +1,12 @@
+{
```
micahcarroll: Remove all the layout copy files

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -145,12 +145,20 @@ def run_agents(self, agent_pair, include_final_state=False, display=False, displ
...
+            if agent_pair.a0.return_action_probs and agent_pair.a1.return_action_probs:
+                a_t, a_info_t = zip(*joint_action_and_infos)
+                assert all(type(a_info) is dict for a_info in a_info_t)
+                info["agent_infos"] = a_info_t
+            elif not agent_pair.a0.return_action_probs and not agent_pair.a1.return_action_probs:
+                a_t = joint_action_and_infos
+
+            assert all(a in Action.ALL_ACTIONS for a in a_t)
             s_tp1, r_t, done, info = self.step(a_t)
-            info["agent_infos"] = a_info_t
```
micahcarroll: Revert back all these changes. Should work fine if you always assume that you will receive a (action, dictionary) pair from your agents's `action` method

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -190,6 +198,7 @@ def get_rollouts(self, agent_pair, num_games, display=False, final_state=False,
 
             # With shape (n_episodes, ):
             "ep_returns": [], # Sum of sparse rewards across each episode
+            "ep_returns_shaped": [], # Sum of dense and sparse rewards across each episode
```
micahcarroll: Similarly revert this line

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -211,6 +220,7 @@ def get_rollouts(self, agent_pair, num_games, display=False, final_state=False,
             trajectories["ep_dones"].append(dones)
             trajectories["ep_infos"].append(infos)
             trajectories["ep_returns"].append(tot_rews_sparse)
+            trajectories["ep_returns_shaped"].append(tot_rews_sparse + tot_rews_shaped * reward_shaping)
```
micahcarroll: Also here

overcooked_ai_py/mdp/overcooked_interactive_vpk.py
```diff
@@ -0,0 +1,270 @@
+import pygame
```
micahcarroll: Again I would probably put in your own project repo, unless it's cleaned up a lot and doesn't require mentions to anything outside of the `overcooked_ai` repo. Same thing about in-depth-reviewing applies here (but again note that I will do more superficial reviews for non-critical code in your project repo).

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1174,6 +1196,172 @@ def get_deltas_to_closest_location(self, player, locations, mlp):
     # DEPRECATED #
     ##############
 
+    """---Added by pk: But these aren't needed AND seem to use an old version of the state encoding, 
+    because sometimes state.objects == state_from_obs_from_state.objects...???"""
+    # def check_state_to_obs_to_state(self, observation, player_observing, state):
+    #
+    #     state_from_obs_from_state = self.state_from_observation(observation, player_observing)
+    #     # print(state_from_obs_from_state)
```
micahcarroll: Remove all of these commented lines. Maybe keep a copy in a local `old` folder, but .gitignore it out of the actual commit.

Did you mean to say  `state.objects != state_from_obs_from_state.objects`? That's not a very good sign, but yeah it is probably due to minor modifications made to the featurization function, and fortunately we don't need that function anymore.

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -884,6 +884,19 @@ def _check_valid_state(self, state):
         for obj_state in all_objects:
             assert obj_state.is_valid()
 
+    def find_free_counters_valid_for_both_players(self, state, mlp):
+        """Finds all empty counter locations that are accessible to both players"""
+        one_player = state.players[0]
+        other_player = state.players[1]
```
micahcarroll: Nicer to write `one_player, other_player = state.players` in one line (tuples automatically unpack)

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -884,6 +884,19 @@ def _check_valid_state(self, state):
...
+        free_counters = self.get_empty_counter_locations(state)
+        free_counters_valid_for_both = []
+        for free_counter in free_counters:
+            goals = mlp.mp.motion_goals_for_pos[free_counter]
+            if (True in [mlp.mp.is_valid_motion_start_goal_pair(one_player.pos_and_or, goal) for goal in goals]) and \
```
micahcarroll: `any([mlp.mp.is_valid_motion_start_goal_pair(one_player.pos_and_or, goal) for goal in goals])` achieves the same thing as `True in ...` and is more elegant

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1071,25 +1084,33 @@ def process_for_player(primary_agent_idx):
...
-        return final_obs_for_players
+        final_obs_p0 = process_for_player(0)
+        final_obs_p1 = process_for_player(1)
+
+        # Check that the func to turn the obs back into a state is working:
+        # self.check_state_to_obs_to_state(final_obs_p0, player_observing=0, state=overcooked_state)
+        # self.check_state_to_obs_to_state(final_obs_p1, player_observing=1, state=overcooked_state)
+
+        return final_obs_p0, final_obs_p1
+
```
micahcarroll: Revert back to original code, unless the obs -> state code is cleaned up sufficiently (but this doesn't seem useful / high priority enough to actually do, given that we're not actually using it)

overcooked_ai_py/planning/planners.py
```diff
@@ -811,19 +811,27 @@ def get_medium_level_actions(self, state, player, waiting_substitute=False):
...
-    def pickup_onion_actions(self, state, counter_objects):
-        onion_dispenser_locations = self.mdp.get_onion_dispenser_locations()
-        onion_pickup_locations = onion_dispenser_locations + counter_objects['onion']
+    def pickup_onion_actions(self, state, counter_objects, only_use_dispensers=False):
+        """If only_use_dispensers is True, then only take onions from the dispensers"""
+        if not only_use_dispensers:
+            onion_dispenser_locations = self.mdp.get_onion_dispenser_locations()
+            onion_pickup_locations = onion_dispenser_locations + counter_objects['onion']
+        else:
+            onion_pickup_locations = self.mdp.get_onion_dispenser_locations()
```
micahcarroll: Don't need to repeat the `self.mdp.get_onion_dispenser_location()` call twice, can place it before the `if` statement and have no `else`.

overcooked_ai_py/planning/planners.py
```diff
@@ -811,19 +811,27 @@ def get_medium_level_actions(self, state, player, waiting_substitute=False):
...
-    def pickup_dish_actions(self, state, counter_objects):
-        dish_dispenser_locations = self.mdp.get_dish_dispenser_locations()
-        dish_pickup_locations = dish_dispenser_locations + counter_objects['dish']
+    def pickup_dish_actions(self, state, counter_objects, only_use_dispensers=False):
+        """If only_use_dispensers is True, then only take dishes from the dispensers"""
+        if not only_use_dispensers:
+            dish_dispenser_locations = self.mdp.get_dish_dispenser_locations()
+            dish_pickup_locations = dish_dispenser_locations + counter_objects['dish']
+        else:
+            dish_pickup_locations = self.mdp.get_dish_dispenser_locations()
```
micahcarroll: Same as above

overcooked_ai_py/mdp/overcooked_interactive_vpk.py
micahcarroll: (I had moved this file out of overcooked_ai a while ago, maybe you still had a old version here)

overcooked_ai_py/agents/agent.py
paulk444: The logging.info calls are there so we can see what the agents are "thinking" as they play, by setting "logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)" when viewing the agents (e.g. in view_agent.py). This is useful for debugging etc because at any timestep we can see what decision the agent is making any why.

Anyway, I guess that once I've finished making the new TOM model (and once I'm sure it's working correctly), then these logging.info calls won't be necessary, so I think it will be fine for me to remove them. Although if any later changes are made to the model then they might be useful again.

I can see that they made the code more messy, so would you advise that I remove all the logging.info calls as soon as I think the new TOM model is working okay?

overcooked_ai_py/agents/agent.py
paulk444: I've now moved my agents (GreedyHumanModel_pk and ToMModel) to my repo: human_ai_robustness.agents. This means that there isn't much depreciated/commented out code left in overcooked_ai_py/agents/agent.py, so it's much cleaner now :)

overcooked_ai_py/agents/view_agent.py
paulk444: hehe oh dear, this code of mine is exceptionally bad!! I've moved it to my own repo now (and also sorted it out!)

overcooked_ai_py/data/layouts/random0.layout
paulk444: One reason for me keeping the original layout names (random0, simple, etc) was because the layout names need to match the names stored in the data for the BC model. I'm using the BC data stored here: https://github.com/HumanCompatibleAI/human_aware_rl/tree/master/human_aware_rl/data/bc_runs Alternatively, I could use the newer data in your human_model_theory repo here: https://github.com/HumanCompatibleAI/human_model_theory/tree/micah_dev/human_model_theory/data/bc_runs

The latter data has the correct layout names (as far as I know), but I've been doing all my benchmarking with the former data so I was reluctant to change. Also, the new data has both "overfitted" and "underfitted" versions for each map.

One option is that I keep using the old data, and keep using the old layout names everywhere except when I actually load the layout, in which case I'll just convert the old names to the new names. This would allow to remove all the layout copy files.

Another option is that I start using the new BC data, and then I can use the new layout names everywhere.

Let me know which you think is the better options. And apologies if I've misunderstood this. Thanks!

micahcarroll: I think the fastest thing to do is to currently do this:

> One option is that I keep using the old data, and keep using the old layout names everywhere except when I actually load the layout, in which case I'll just convert the old names to the new names. This would allow to remove all the layout copy files.

And try to keep this hack easily removable once the naming issue is fixed. It's quite hacky, but I feel like having duplicates in overcooked_ai is worse.

overcooked_ai_py/agents/agent.py
micahcarroll: Yeah it's fine to keep the logging.info calls in your own project repo for now (given that you moved the agent, it's actually fine), was just reluctant to committing them to overcooked_ai

overcooked_ai_py/mdp/overcooked_env.py
paulk444: I like your solution to this ("define a metadata_fn that can extract the shaped_reward information...") -- very neat :)

overcooked_ai_py/agents/agent.py
paulk444: Yep fair enough

overcooked_ai_py/data/layouts/random0.layout
paulk444: I've now removed the duplicate layouts. I tried a few methods to make my changes as non-hacky as possible. But the best I came come up with was this:
- I added a function to utils that converts the layout names *if* an old name is used (so if a new name is used then this function won't do anything): https://github.com/HumanCompatibleAI/human_ai_coord/blob/c30dc0bb01a9f504915df6703b0c5f72f1ae1d3d/human_aware_rl/utils.py#L138
- I then made a few changes to my own code (pbt_hms and ppo_tom) to make things work. The only change I made to the general code was to behavioural_cloning, here: https://github.com/HumanCompatibleAI/human_ai_coord/blob/c30dc0bb01a9f504915df6703b0c5f72f1ae1d3d/human_aware_rl/imitation/behavioural_cloning.py#L91

PR #22, opened 2020-04-24
Adding new changes from micah_dev to improvements branch

micahcarroll:
Things done in this PR:

`overcooked_ai_js`:
- Added common tests across JS and PY for lossless state encoding
- Added infrastructure to easily add featurization encoding common test
- Improved common trajectory tests framework
- Made naming of JS package consistent ("overcooked" from "overcook"), this might require uninstalling/reinstalling the JS package
- Temporarily discontinued support of tomatoes in JS
- Did some refactoring / formatting

`overcooked_ai_py`:
- Fixed a couple of minor bugs in the environment code (that affected featurizations – but did not affect DRL results in any perceptible way)
- Added events tracking in `overcooked_mdp`: now various types of events (onion/dish/soup pickups/drops are logged and will be returned together with the trajectory info)
- Added tracking of reward contribution by agent (in infos reward is now also broken down by which agent actually contributed to obtaining it). This can be helpful for debugging DRL runs / figure out if an agent is actually doing anything useful (one can see this breakdown both for sparse and shaped rewards). 
- Moved the HH data to this repo
- Added support for the environment to start randomly from HH data states (make sure this is standalone)
- Improved common trajectory tests framework
- Made all agents require support for the `actions(states)` method, that enables to more easily upgrade get_rollouts to be parallelization friendly (this helps especially with NNs)
- Improved interactive trajectory visualization for Jupyter Notebooks
- Improved layout importing background code
- Changed terminal graphics to always display agent index

One issue that I probably should address before merging:
- Currently although I moved the unparsed HH data is in this repo (in a pandas dataframe), the parsing functions are in human_aware_rl. I should probably also add a cleaned & already python-converted version to this repo, and call it a day. If I really wanted to do things right, I should move part of the human data parsing code to this repo (the one that is HH-data specific), but I'm not sure it's worth it currently.


micahcarroll:
Another thing I should probably do before merging is having some utils to visualize final agent behavior (in addition to the utils I built in human_ai_coord to analyze it over the course of training)

overcooked_ai_py/agents/agent.py
```diff
@@ -119,21 +120,22 @@ def joint_action(self, state):
...
     Defines an agent from a `state_policy` and `direct_policy` functions
     """
-
-    def __init__(self, state_policy, direct_policy, stochastic=True, action_probs=False):
+    
+    def __init__(self, state_policy, direct_policy, multi_state_policy, stochastic=True, action_probs=False):
         """
         state_policy (fn): a function that takes in an OvercookedState instance and returns corresponding actions
         direct_policy (fn): a function that takes in a preprocessed OvercookedState instances and returns actions
+        multi_state_policy (fn): a function that takes in multiple OvercookedState instatences and returns actions
```
paulk444, line 132: Minor typo here! instatences --> instances
(I'm not sure if I should flag this in PR review?!)

paulk444: (This multi_state_policy is a good idea by the way)

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -18,11 +18,14 @@ class OvercookedEnv(object):
...
     """
 
-    def __init__(self, mdp, start_state_fn=None, horizon=MAX_HORIZON, debug=False):
+    def __init__(self, mdp, start_state_fn=None, horizon=MAX_HORIZON, debug=False, hh_data_starts=False):
         """
         mdp (OvercookedGridworld or function): either an instance of the MDP or a function that returns MDP instances 
         start_state_fn (OvercookedState): function that returns start state for the MDP, called at each environment reset
         horizon (float): number of steps before the environment returns done=True
+
+        TODO: make the start state completely decided in the MDP. MDPs with different start states should be considered
```
paulk444, line 27: What would be the reason for doing this? Making a new MDP can take a while for some layouts, so on the rare occasion that lots of different start states need to be used (such as when doing the qualitative tests!) then having to make a new MDP each time might slow things down.

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+        return no_dishes_on_counters and num_player_dishes < non_empty_pots
+
+    def is_dish_drop_useful(self, state, pot_states, player_index):
+        """
+        NOTE: this only works if self.num_players == 2
+        Useful if:
+        - Onion is needed (all pots are non-full)
+        - Nobody is holding onions
+        """
+        if self.num_players != 2: return False
```
paulk444, line 747: Shouldn't this give an error if there aren't 2 players, rather than returning False? (I.e. "assert self.num_players == 2"

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+        if all_pots_full and not other_player_has_dish:
+            return False
+        return True
+        
+    def is_onion_drop_useful(self, state, pot_states, player_index):
+        """
+        NOTE: this only works if self.num_players == 2
+        Useful if:
+        - Dish is needed (all pots are full)
+        - Nobody is holding a dish
```
paulk444, line 772: There could be some rare cases that don't fit into the criteria for these various events being useful. E.g. An onion drop can be useful if the other player has an onion and is closer. But I don't think these rare cases would change the overall stats much so they probably don't matter.

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -33,6 +36,22 @@ def __init__(self, mdp, start_state_fn=None, horizon=MAX_HORIZON, debug=False):
         else:
             raise ValueError("Mdp should be either OvercookedGridworld instance or a generating function")
         
+        if hh_data_starts:
+            assert start_state_fn is None, "Either you have hh_data_starts, or you set a custom start_state_fn"
```
rohinmshah, line 40: I assume it's fine to have neither? In that case it should be `"Cannot have both hh_data_starts and a custom start_state_fn"`

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -134,7 +179,8 @@ def run_agents(self, agent_pair, include_final_state=False, display=False, displ
         """
         Trajectory returned will a list of state-action pairs (s_t, joint_a_t, r_t, done_t, info_t).
         """
-        assert self.cumulative_sparse_rewards == self.cumulative_shaped_rewards == 0, \
+        assert all([rew == 0 for rew in self.game_stats["cumulative_shaped_rewards_by_agent"]]) and \
+            all([rew == 0 for rew in self.game_stats["cumulative_sparse_rewards_by_agent"]]), \
             "Did not reset environment before running agents"
```
rohinmshah, line 184: Shouldn't it be something more like `assert not self.game_stats[key]`, that is, the game stats should be empty?

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+        """
+        NOTE: this only works if self.num_players == 2
+        Useful if:
+        - Pot is ready/cooking and there is no player with a dish               \ 
+        - 2 pots are ready/cooking and there is one player with a dish          | -> number of dishes in players hands < number of ready/cooking/partially full soups 
+        - Partially full pot is ok if the other player is on course to fill it  /
+
+        We also want to prevent picking up and dropping dishes, so add the condition
+        that there must be no dishes on counters
+        """
```
rohinmshah, line 731: This is missing the check for `num_players` being 2

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+        Always useful unless:
+        - All pots are full & other agent is not holding a dish
+        """
+        if self.num_players != 2: return False
+        all_pots_full = self.num_pots == len(self.get_full_pots(pot_states))
+        other_player = state.players[1 - player_index]
+        other_player_has_dish = other_player.has_object() and other_player.get_object().name == "dish"
+        if all_pots_full and not other_player_has_dish:
+            return False
+        return True
```
rohinmshah, line 765:
```suggestion
        return not (all_pots_full and not other_player_has_dish)
```

overcooked_ai_py/mdp/overcooked_mdp.py
rohinmshah: I think the point is to allow the code to run with > 2 players, but just not collect these statistics in those cases.

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+            events_infos["onion_pickup"][player_index] = True
+            if self.is_onion_pickup_useful(state, pot_states, player_index):
+                events_infos["useful_o_pickup"][player_index] = True
+
+        elif obj_name == "tomato":
+            # TODO: logging/events for tomatoes are not currently supported
+            pass
+        
+        else:
+            raise ValueError()
```
rohinmshah, line 808:
```suggestion
        obj_pickup_key = obj_name + "_pickup"
        if obj_pickup_key not in events_infos:
            raise ValueError("Unknown event {}".format(obj_pickup_key))
        events_infos[obj_pickup_key][player_index] = True
        
        USEFUL_PICKUP_FNS = {
            "onion": self.is_onion_pickup_useful,
            "dish": self.is_dish_pickup_useful
        }
        if obj_name in USEFUL_PICKUP_FNS:
            if USEFUL_PICKUP_FNS[obj_name](state, pot_states, player_index):
                obj_useful_key = "useful_" + obj_name + "_pickup"
                events_infos[obj_useful_key][player_index] = True
```

Note this requires you to change the key names from `useful_o_pickup` to `useful_onion_pickup` (and same for dishes), and also requires you to add an unused `player_index` argument to `is_dish_pickup_useful`.

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -669,27 +696,151 @@ def get_state_transition(self, state, joint_action):
...
+            events_infos["onion_drop"][player_index] = True
+            if self.is_onion_drop_useful(state, pot_states, player_index):
+                events_infos["useful_o_drop"][player_index] = True
+
+        elif obj_name == "tomato":
+            # TODO: logging/events for tomatoes are not currently supported
+            pass
+
+        else:
+            raise ValueError()
```
rohinmshah, line 830:
```suggestion
        """Player dropped the object on a counter"""
        obj_drop_key = obj_name + "_drop"
        if obj_drop_key not in events_infos:
            raise ValueError("Unknown event {}".format(obj_drop_key))
        events_infos[obj_drop_key][player_index] = True
        
        USEFUL_DROP_FNS = {
            "onion": self.is_onion_drop_useful,
            "dish": self.is_dish_drop_useful
        }
        if obj_name in USEFUL_DROP_FNS:
            if USEFUL_DROP_FNS[obj_name](state, pot_states, player_index):
                obj_useful_key = "useful_" + obj_name + "_drop"
                events_infos[obj_useful_key][player_index] = True
```

overcooked_ai_py/mdp/overcooked_mdp.py
micahcarroll: Yup, Rohin is right

overcooked_ai_py/mdp/overcooked_env.py
micahcarroll: I thought it would be a good idea because it makes it more compatible with AI Theory: an MDP is defined as having, among other things, a start state distribution. So just to make it more intuitive to people that know the theory, it might be nicer to keep the abstractions consistent with the theory.

Code wise, it also seems that dealing with start states in the mdp might make things a little bit more simple (but it does admittedly increase the burden on the MDP class, that is already quite big)

overcooked_ai_py/agents/agent.py
micahcarroll: Yeah, I've actually changed this API a tiny bit more, will be in a newer commit

overcooked_ai_py/mdp/overcooked_mdp.py
micahcarroll: Yeah, I think the "useful" stats should all be taken as approximate/with a grain of salt. I mainly want to use it as a diagnostic of "if i see that <20% of X drops are useful", that's probably a bad sign / a sign that i should check what is going on

micahcarroll:
Paul: 
the thoroughness depends a bit on your own preference (I don't think I have a particularly strong one either way). The comments you left seemed to be on the level of abstraction I was expecting, so I think this is good! Re: the contribution reward, I was actually recently speaking with MSR about this, and might try messing with that in the near future.

Rohin: 
> we could clean up the implementation of the game dynamics
Do you mean the overcooked_mdp file? Yeah, that is on my todo list, it is becoming quite a mess 😅

overcooked_ai_py/mdp/overcooked_env.py
micahcarroll: Yeah this was a hacky way of doing it that never got changed, will update

micahcarroll:
I've moved the hh data starts functionality to the human_ai_theory repo, as it resolves various style issues and is more clean compared to the current setup. I'm planning on merging `overcooked_ai_improvements` to `master` after this, as the improvements branch contains a lot of important changes that I think should be reflected sooner rather than later (we've recently had various new stars on the repo for some reason)

overcooked_ai_py/mdp/overcooked_env.py
micahcarroll: Actually to be fair, if one wants to be pedantic about conforming to theory, we should also separate VariableMDPs into their own MDP class (and not have that handled by the OvercookedEnv), as VariableMDPs are themselves a specific type of MDP in which there is a distribution over decision process structure.

This might make sense to do eventually, as the variable MDP distinction is a bit annoying to deal with within the Env code, and introduces some issues around the place. But in light of this, it also seems plausible that both this and the start state changes would complicate things more than they simplify them.

paulk444: Ah okay, this makes sense!

overcooked_ai_py/mdp/overcooked_mdp.py
paulk444: Yeh that sounds right

paulk444:
> ..., re: training with the agent contribution reward, I was actually recently speaking with MSR about this, and might try messing with that in the near future.

Cool, I'd be interested to hear how this goes!

PR #32, opened 2020-06-24
overcook_ai improvements for multi layout training

mesutyang97:
Main changes are in the AgentEvaluator class, to work with the multi-layout Overcooked environment (accommodate mdp_params_lst)

Also added cramped_room_2, another layout with the same size as cramped_room.

Need to work with the _multi_layout_train_ commit on the **human_aware_rl** repo.


overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,46 +20,71 @@ class AgentEvaluator(object):
     pickleable. We should think about possible improvements/what makes most sense to do here.
     """
 
-    def __init__(self, mdp_params, env_params={}, mdp_fn_params=None, force_compute=False, mlp_params=NO_COUNTERS_PARAMS, debug=False):
+    def __init__(self, mdp_params_lst, env_params={}, mdp_fn_params_lst=None, force_compute=False, mlp_params_lst=None, debug=False):
```
micahcarroll, line 23: Just checking, because it's not obvious to me – although I might be missing something: is there a reason here that we need mdp_params_lst? 

Couldn't we have just have 1 mdp_params dictionary, which has a mdp generator function that returns a different mdp at each reset?

Having a list everywhere makes things quite a bit more redundant notation-wise for all the cases in which one deals with a single layout (which is probably most of the time).

Moreover, if you're planning on training on infinitely many randomly generated layouts, you won't be able to enumerate them here in any case?

overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,46 +20,71 @@ class AgentEvaluator(object):
...
         else:
-            self.mdp_fn = LayoutGenerator.mdp_gen_fn_from_dict(mdp_params, **mdp_fn_params)
-            self.env = OvercookedEnv(self.mdp_fn, **env_params)
+            assert type(mdp_fn_params_lst) is list and len(mdp_fn_params_lst) > 0
+            assert all([type(mdp_fn_params_i) is dict for mdp_fn_params_i in mdp_fn_params_lst]), \
+                "all mdp_fn_params must be a dictionary"
+            for mdp_params_i, mdp_fn_params_i in zip(mdp_params_lst, mdp_fn_params_lst):
+                mdp_fn = LayoutGenerator.mdp_gen_fn_from_dict(mdp_params_i, **mdp_fn_params_i)
+                self.mdp_fn_lst.append(mdp_fn)
+                self.env_lst.append(OvercookedEnv(mdp_fn, **env_params))
```
micahcarroll, line 53: Eventually here you are basically doing the same thing as was done before (using LayoutGenerator for each mdp_params dict), but all this functionality could be embedded in the LayoutGenerator and the mdp_fn_params I believe?

overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,46 +20,71 @@ class AgentEvaluator(object):
...
-            mdp = OvercookedGridworld.from_layout_name(**mdp_params)
-            self.mdp_fn = lambda: mdp
-            self.env = OvercookedEnv.from_mdp(mdp, **env_params)
+        assert type(mdp_params_lst) is list and len(mdp_params_lst) > 0, \
+            "please make sure mdp_params_lst is a list. " \
+            "If not, please see where the function is called and make a fake list"
+        assert all([type(mdp_params_i) is dict for mdp_params_i in mdp_params_lst]), \
+            "all mdp_params must be a dictionary"
+
+        # The main changes here is that we are putting a list of environment in the AgentEvaluator.
```
micahcarroll, line 38: Just to clarify, what I'm suggesting is to always have 1 Environment object that is capable of dealing with all the complexity of multiple layouts within its code (i.e. same environment object has different mdps at each reset, rather than having a bunch of fixed-layout Envs).

overcooked_ai_py/agents/benchmarking.py
micahcarroll: Just for some more context: my reasoning behind this reticence to modify this signature is that, because the Env class is so widely used across the codebase, it seems good to try and change it as little as possible, unless there's a strong case for why the change is absolutely necessary here or is clearly stylistically superior and will result in less bugs and cleaner code in the future.

overcooked_ai_py/agents/benchmarking.py
micahcarroll: If it's possible to do this elegantly, it will pay off even for the rllib training itself (for which, again, we will only need 1 environment rather than keeping track of many).

overcooked_ai_py/agents/benchmarking.py
mesutyang97: the reason we have a list here is because we don't necessarily have to train on environment that is coming from the same exact distribution (coming out of the same set of mdp_params). Having a list enable us to train the same agent simultaneously on two very different layout (in terms of both layout and distribution of objects)

And for the second question: we don't have to enumerate over all randomly generated layouts if we want to train on infinitely many of them because once the layout_generator works properly, we can generate it from a parameters directly instead of using "layout_name"s

micahcarroll:
First off, I'm pretty impressed that you could wrap your head around the codebase enough to make these changes so quickly! 😊

But yeah, as I mentioned in the comments, it would probably be nice to find a different way to deal with multi-mdp Environments. I had implemented multiple-layout training in the past (but then discontinued it once we decided not to look into that direction yet), and seem to remember it not being necessary to modify the Env. In any case, lmk what you think!

overcooked_ai_py/agents/benchmarking.py
mesutyang97: Agreed. I noticed the similarity, but I just strictly followed the original code for this version. I will make an update to it after I fix the LayoutGenerator.

overcooked_ai_py/agents/benchmarking.py
mesutyang97: The main problem with having 1 environment object is that I really cannot find a way to "step" in 1 environment but multiple layouts.

I read through the relevant files in rllib, and tried that for a pretty long time and had no results.

I think I will work on the LayoutGenerator first and come back to see if I can amend this

mesutyang97:
I agree we won't need to have a list of mdp_params, if all the layout is coming from the same distribution. But I thought it would be nice to see how a learnt model can generalize to layouts of not just different geometry, but different properties (which will require a different set of mdp_params)

mesutyang97:
But I think I can implement what you described without using a list. I can start off with that as well. Thanks for all the feedback!

micahcarroll:
Yeah I think that trying to make it work with different properties makes sense (I wasn't currently planning on using something like that, but i feel like it's plausible that we would want to use it in the future) – for example, we could have the mdp_generating function generate an entire mdp _with_ mdp_params embedded in it.

However, we could try to add this later on, as i think it would possibly require some more extensive restructuring than the basic change. Up to you, depending on how much work it would be.

overcooked_ai_py/agents/benchmarking.py
micahcarroll: Hmm I see – in the overcooked_ai portion of the code that should be relatively straightforward and it worked fine for baselines, but yeah, I'm not sure if it would cause complications for rllib (although i would be surprised if one couldn't get around it).

mesutyang97: rllib sealed off a lot of functionality: I actually cannot even edit the train function in rllib

After poking around for a long time, I realize the only API we have control over is via the step function in the OvercookedEnv. Techinically that is still only 1 OvercookedEnv with many layout_env inside.

mesutyang97:
OK, so should I keep mdp_params list or should I revert it? I am biased towards keeping the list, because we technically still only have 1 outside Environment, and the list is giving us the possibility for different mdp properties, it leaves the option to tune the diversity of training data

micahcarroll:
I feel like it would be unnecessarily burdensome in terms of notation to have the list. If we want to have different mdp_params across different mdps, we can just have the mdp_generator generate an mdp_param dict too, together with the mdp. 

One more thought: one simplifying thing that we should do is to change the signature of AgentEvaluator to have mdp_fn in the __init__ method, and then have a staticmethod called `from_mdp_params` that creates a dummy mdp_fn that simply returns the same mdp each time.

mesutyang97:
I actually am not very sure why it would be unnecessary to include the list in the notation. If we are training a network that can play on 2 layouts simultaneously, we are indeed optimizing over 2 MDP, is that correct?

The main problem is that with rllib, we have no access to the train function (unless we decide to modify the rllib trainer code in the library). So having a list of envinronment is a workaround.

We can have just 1 set of mdp_params, but I believe it is necessary to keep 2 mdps (and thus 2 virtual OvercookedEv) in the Overcooked environment, because the 2 mdps will have to have different featurize_fn and different mlp too.

mesutyang97:
We can do from_mdp_params.

The if statements are from the old code. I think the if statement in _init_ can be naturally fixed after LayoutGenerator is fixed.

micahcarroll:
We should probably close this PR given that #49 basically has most of the changes here + more! @mesutyang97

PR #35, opened 2020-07-27
Tutorial

nathan-miller23:
### Layouts
* added `tutorial_{0, 1, 2, 3}.layout` files to be used for the /tutorial route in overcooked-demo

### Bug fixes
* Modified  the `OvercookedState.from_dict` function to properly de-serialize `bonus_orders` (serialization worked before but de-serialization failed)
* Modified the `SoupState.from_dict` function to properly de-serialize 
* Modified the `Recipe.from_dict` function to fix de-serialization
* Modified the `SoupState.to_dict` function to fix silly bug (I had `stat_dict['is_ready'] = self.is_idle` and `state_dict['is_idle`] = self.is_ready` before)

### New features
* Made the `Recipe` class comparable
  * Before, `all_orders` and `bonus_orders` were sets, and thus serialization order was non-deterministic
  * This causes the orders to "flicker" in `Overcooked-Demo`, as the display order rapidly shifted on state updates
   * Now, both `all_orders` and `bonus_orders` are lists, and they are sorted when the state is serialized

* Added optional `start_state` parameter to the `OvercookedGridworld` constructor
  * You can pass in a serialized state-dict to the `start_state` and it will override the state returned by `get_standard_start_state`
  * This was to allow me to have a tutorial layout that begins with a soup already cooking, and it seemed like a nice feature to have in general

### DIFF
--- a/overcooked_ai_py/data/layouts/tutorial_0.layout
+++ b/overcooked_ai_py/data/layouts/tutorial_0.layout
@@ -6,6 +6,6 @@
                 XXXXXXX""",
     "start_all_orders" : [
         { "ingredients" : ["onion", "onion", "onion"]}
-    ]
+    ],
     "rew_shaping_params": None
 }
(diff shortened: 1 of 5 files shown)


micahcarroll:
LGTM – I like the new tutorial! We might want to later also add different cooking times, but I don't think that it will affect how we structure the tutorial itself!

PR #38, opened 2020-08-04
Trajectory visualization

bmielnicki:
Changes are mostly for method events_visualization in AgentEvaluator from overcooked_ai_py/agent/benchmarking.py
Example look:
chart-with-legends

```
@staticmethod
def events_visualization(trajs, traj_index=0, ipython=False, chart_settings=None):
    """
    Displays chart with visualization of events (when items pickups happened etc.)
    ipython - chooses between opening chart
    in default browser or below ipython cell (for ipython=True)
    chart_settings - json with various chart settings that overwrittes default ones
    for more info see create_chart_html function with comments aboutr default chart settings
    """
```

More on chart settings and the way defaults are updated:
```
def create_chart_html(events_data, chart_settings=None):
    # all numerical values are pixels
    # not explicity stated below margins are not implemented
    default_chart_settings = {'height': 250, # height of whole chart container
        'width': 720,  # width of whole chart container
        'margin': {'top': 20, 'right': 60, 'bottom': 180, 'left': 40}, # margins of chart (not container) without legends
        'hold_line_width': 3, # hold line is line between pickup and drop event that symbolizes holding item by the player
        'highlighted_hold_line_width': 6, # highlighting element is by hovering mouse over something associated with described object
        'object_line_width': 0, 'highlighted_object_line_width':3,
        'object_event_height': 10, 'object_event_width':16, # object event is triangle
        'label_text_shift': 30, # how far from axis is axis label
        'add_cumulative_data': True, # cumulative data is data about cumulative events
        'cumulative_data_ticks': 4,
        'show_legends': True, # when setting show_legends to False it is recommended to also lower height and margin.bottom
        'legend_title_size': 10, 'legend_points_height': 10, 'legend_points_width': 16, 
        'legend_points_margin': {'bottom': 5, 'right': 5}, # margin after legend point and next point or legend text
        'legend_margin': {'right':  5} # margin between legend columns
        }

    settings = default_chart_settings.copy()
    settings.update(chart_settings or {})
```

Example usage looks mostly like that:
```
from overcooked_ai_py.agents.benchmarking import AgentEvaluator
agent_eval = AgentEvaluator({"layout_name": "cramped_room"}, {"horizon": 100})
trajectory_human_pair = agent_eval.evaluate_human_model_pair(num_games=1, display=False)
agent_eval.events_visualization(trajectory_human_pair, traj_index=0)
```

There is also small tutorial notebook I wrote: overcooked_ai/overcooked_ai_py/tutorial_notebooks/Agent_evaluator_intoduction.ipynb .

Some of the things I'm not sure about:
- Changing items (e.g. filling dish with soup) is treated as both pickup and drop (drop dish, pickup soup) what increments cumulative stats by 2.
- I put all helper functions that won't be probably used outside extract_events inside the function. I can move them outside of the function (especially bigger ones), but I do not want to suggest they have use outside of the function.
- The function of extraction events is adding ids to object as post-processing of trajectory (to highlight single object events feature). While now this works perfectly its less robust to modifications (e.g. code is assuming that object on the ground won't move without a player, filling dish with soup creates new object_id for soup instead of taking object_id of the dish) than putting ids of objects in MDP files so it is temporary. I can add object ids to original objects when I will change MDP files.
- Using d3.js and require.js from URL, so it won't work offline. I wanted to do it in a more elegant way (from an offline package), but after a couple of failed tries I gave up (I do not want to spend too much time on this detail if it's not important).
- All CSS and js code used in the chart is copied straight into HTML - this ensures HTML with chart will always work (e.g. will be no CORS errors), on the other hand, this adds 7 KiB to every HTML with the chart and looks less elegant than importing files.
- I upped canvas version to 2.6 because installation of canvas 2.5 throw errors (tries to download a file that does not exist, this is the problem with npm package).
- I've moved all visualization code to another folder (accessible from the AgentEvaluator object method) because we probably want to avoid the object of size 1k + lines. I'm not sure if this folder should exist in agents folder or directly below overcooked_ai_py.


micahcarroll:
Hi Bartek, thanks for you PR!

Just as a quick note, it would generally be useful if you add an example usage of the new functionality you introduce in your PRs descriptions (if it's not obvious), and in the case of visualizations also add an example jpeg!

In any case, I do think that it seems to me like many of the changes here are not necessarily compatible in philosophy with the updated repo, so we might just want to leave these changed pending and not merge this PR (I'm sure at a later time, some of this code could be salvaged). In particular, we have now moved to a purely-python implementation of `overcooked_ai` (using the Flask app in `overcooked-demo`, which has only the most essential js code) – @nathan-miller23 (who set this new system up) can correct me if I'm saying something wrong here. This seems at odds with having the visualizations depend on a js module, instead of using purely python-based visualization methods (such as matplotlib). However, I haven't looked too in depth into your changes, so I might also be missing something.

Also, not sure if you saw that this branch is out-of-date with the base branch, so in any case it would require updating before it would even be possible to merge.

bmielnicki:
> Hi Bartek, thanks for you PR!
> 
> Just as a quick note, it would generally be useful if you add an example usage of the new functionality you introduce in your PRs descriptions (if it's not obvious), and in the case of visualizations also add an example jpeg!
> 
> In any case, I do think that it seems to me like many of the changes here are not necessarily compatible in philosophy with the updated repo, so we might just want to leave these changed pending and not merge this PR (I'm sure at a later time, some of this code could be salvaged). In particular, we have now moved to a purely-python implementation of `overcooked_ai` (using the Flask app in `overcooked-demo`, which has only the most essential js code) – @nathan-miller23 (who set this new system up) can correct me if I'm saying something wrong here. This seems at odds with having the visualizations depend on a js module, instead of using purely python-based visualization methods (such as matplotlib). However, I haven't looked too in depth into your changes, so I might also be missing something.
> 
> Also, not sure if you saw that this branch is out-of-date with the base branch, so in any case it would require updating before it would even be possible to merge.


I think both pure python and python+js approaches for visualizations are ok depending on the interactivity level we want to have.

Pros of having charts in js runnable from python:
- easy to make interactive charts (now used only for highlighting events associated with a single item, this alone probably does not justify adding js)
- easy to add another js code like a graphic-based version of AgentEvaluator.interactive_from_traj in the same HTML

Cons of having charts in js:
- current charting code requires both python (pre-processing trajectory data) and javascript (representation)
- need some workarounds to work in a web browser (like saving HTML in /tmp directory to open it from file)

Additional questions:
1. Do we want to add a graphic-based version of AgentEvaluator.interactive_from_traj? If yes, do we want (and how) use code from overcooked-demo/server/graphics/overcooked_graphics_v2.1.js ?
2. Should I refactor code now (migrating some code to MDP from post-processing of trajectory, maybe migrating js code to pure python if it will be confirmed as the way to go) or wait with it? It mostly depends if we want to merge it in the near future to master (if not I think it's better to wait with it, many stuff can change in the meantime).

overcooked_ai_py/visualization/extract_events.py
```diff
@@ -0,0 +1,200 @@
...
+                    if held_object.get("name") != held_object_next_t.get("name"):
+                        add_drop_item_event(events, player, p, t, held_object, terrain)
+                        add_holding_event(events, p, t, held_object)
+                
+                if held_object_next_t and held_object.get("name") != held_object_next_t.get("name"):
+                    add_pickup_item_event(events, player, p, t, held_object_next_t, terrain)
+                    
+            else: #last timestep
+                if held_object:
+                    add_holding_event(events, p, t, held_object)
```
nathan-miller23, line 185: This code seems redundant and abstraction-barrier breaking -- we already have code for detecting events (such as picking up objects, dropping them, various soup creation and potting events) in the `overcooked_mdp` class. In fact, the way I would recommend doing this is by parsing the `trajectory['ep_infos']['agent_infos'][<EVENT_NAME>]` dictionaries. You can find a full list of events currently supported in the `EVENT_TYPES` global in `overcooked_mdp.py`. This way, the visualization implementation doesn't depend on the exact implementation details of OvercookedState

nathan-miller23: This would also allow for potentially richer visualization; we distinguish between "useful" and "non-useful" events (i.e. picking up a tomato is non-useful when all pots are full) which gives an extra dimension to the trajectory data

overcooked_ai_js/package.json
```diff
@@ -21,7 +21,7 @@
     "babel-core": "^6.26.3",
     "babel-preset-env": "^1.7.0",
     "babelify": "^8.0.0",
-    "canvas": "^2.5.0",
+    "canvas": "^2.6.0",
     "jest": "^23.6.0",
     "jest-canvas-mock": "^2.1.0",
     "jest-expect-message": "^1.0.2",
```
nathan-miller23, line 27: We do plan on completely depricating all of the code in `overcooked_ai_js/`, so your updates shouldn't touch any files in this sub-directory

overcooked_ai_py/agents/benchmarking.py
```diff
@@ -379,8 +381,19 @@ def update(t = 1.0):
...
+        in default browser or below ipython cell (for ipython=True)
+        chart_settings - json with various chart settings that overwrittes default ones
+        for more info see create_chart_html function with comments aboutr default chart settings
+        """
+        events = extract_events(trajs, traj_index)
+        html = create_chart_html(events, chart_settings)
+        if ipython:
+            run_html_in_ipython(html)
+        else:
+            run_html_in_web(html)
```
nathan-miller23, line 399: I would also recommend writing 1-2 unit tests for each new piece of code you write. It's a good habit to get into, and even if it seems silly as the code you'd be testing is very simple or high level, it's a great way to ensure that downstream changes don't break upstream things like visualization tools

nathan-miller23 (dismissed):

overcooked_ai_py/agents/benchmarking.py
bmielnicki: Unit tests added

overcooked_ai_js/package.json
bmielnicki: change reversed

overcooked_ai_py/visualization/extract_events.py
bmielnicki: I've done small refactor of event logging to be more flexible:
- added event_list that contains all of the events as a list of dicts
    - it's not ideal that we have both ep_game_stats and event_list with similar info, on the other hand, I needed more info than in ep_game_stats alone and did not want to delete ep_game_stats/modify it too much
- Added adjectives of events ("useful", "viable" etc.) into extracted events so in future this can be incorporated into charts (if  it would be immediately useful I can do it before other tasks)
- I've added JSON serialization of NumPy arrays. Implemented it only for ep_infos as I'm not sure what "metadatas" field in trajectory contains

src/overcooked_ai_py/visualization/extract_events.py
```diff
@@ -0,0 +1,90 @@
...
+        }
+        return event
+
+    def get_holding_events(events, game_states):
+        holding_events = [holding_event(events, e.get("player"), e.get("timestep"), e.get("object")) 
+            for e in events if e.get("action") in ["drop", "potting", "delivery"]]
+        # check if object is held at end of the episode
+        for p, player in enumerate(game_states[-1].players):
+            if player.held_object:
+                holding_events.append(holding_event(events, p, len(game_states)-1, player.held_object.to_dict()))
```
nathan-miller23, line 35: Still not a fan of how this code has to "look inside" the overcooked mdp class at the game_state dictionary representation in order to function. You don't have to change it right now, because I'm hopeful that adding the trajectory class (as discussed in a different issue) will clean up this code a bunch

testing/visualization_test.py
```diff
@@ -0,0 +1,38 @@
...
+from overcooked_ai_py.visualization.visualization_utils import create_chart_html
+from overcooked_ai_py.agents.benchmarking import AgentEvaluator
+from overcooked_ai_py.utils import load_from_json, NumpyArrayEncoder
+from utils import TESTING_DATA_DIR
+from overcooked_ai_py.mdp.overcooked_mdp import Recipe
+
+def jsonify(obj):
+    return json.loads(json.dumps(obj, cls=NumpyArrayEncoder))
+
+class TestVisualizations(unittest.TestCase):
```
nathan-miller23, line 13: Make sure you add this test to the `/.github/workflows/pythontests.yml` file so it executes on PRs and pushes. While in the future, any file matching `/testing/*_test.py` will get automatically executed, that's not currently the case!

testing/visualization_test.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_create_chart_html(self):
+        box_id = "graph-div-" + str(uuid.uuid1())
+        html = create_chart_html(self.extracted_events1, box_id=box_id)
+
+        self.assertTrue("it is chart.html file" in html)
+        self.assertTrue("it is event_chart.js file" in html)
+        self.assertTrue("it is style.css file" in html)
+
+        self.assertTrue("#"+box_id in html)
+        self.assertFalse("$" in html)
```
nathan-miller23, line 34: I'm curious, why is this a necessary assert? wouldn't adding jquery code to the JS cause this to fail?

overcooked_ai_py/visualization/extract_events.py
nathan-miller23: I left a comment asking for clarification on why you needed the separate `event_list` in addition to the `ep_game_stats` but it appears that you did that to keep track of object IDs. I think it would be perfectly fine to add that to the `ep_game_stats` dictionary -- as a general rule of thumb, adding new fields to existing dictionaries (like a nested 'object_id' field) is fine as long as you don't remove existing fields.

nathan-miller23: for the sake of this PR, however, we can keep all the trajectory parsing as it is, as I'm hoping that the refactor issue, which involves writing a Trajectory class wrapper over the increasingly messy dictionaries, will clean up a lot of this logic

src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1592,16 +1580,16 @@ def log_object_potting(self, events_infos, state, old_soup, new_soup, obj_name,
...
+        for outcome in outcomes:
+            potting_key = "{}_{}_potting".format(outcome, obj.name)
+            events_infos[potting_key][player_index] = True
 
-        for outcome, outcome_fn in POTTING_FNS.items():
-            if outcome_fn(state, old_soup, new_soup):
-                potting_key = "{}_{}_potting".format(outcome, obj_name)
-                events_infos[potting_key][player_index] = True
-
+        events_list.append(dict(action="potting", player=player_index, object=obj.to_dict(), adjectives=outcomes))
```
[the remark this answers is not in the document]
nathan-miller23: Ok I see, it appears it was to differentiate between objects with the same name

testing/visualization_test.py
bmielnicki: This is an assert because I wanted to make sure there is no variable added in the future to replace (so there is no need to update test) that is not replaced as there is currently no other usage for dollar sign in the code. On the other hand, adding jquery code would trigger this assert so I will change it to check variables using "$"+variable_name .

src/overcooked_ai_py/visualization/extract_events.py
```diff
@@ -0,0 +1,90 @@
...
+        # add zero timestep events
+        first_timestep = 0
+        add_cumulative_event(cumulative_events, s, first_timestep, a)
+        for p in set(map(lambda x: x["player"], events)):
+            add_cumulative_event(cumulative_events, s, first_timestep, a, p)
+
+        for e in filter(lambda x:x["action"] in ["pickup", "drop"], events):
+            p = e["player"]
+            t = e["timestep"]
+            s+=1
```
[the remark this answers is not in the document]
bmielnicki: > Overall looks good. I left some pretty nitpicky feedback but all the requested changes should be very minor, if anything at all. As a general comment, I would be very careful with the object_id field -- I like the idea of adding a unique id to each game object, but I can forsee a lot of pitfalls (i.e. a serialized and then deserialized object not keeping its uid) as well as logic pitfalls, as the rest of Overcooked_mdp.py was written without the uid in mind. There might be cases where we are creating new objects that would have previously been considered equal instead of keeping the old object around. For example, before we used to create a new Onion object when it was added to a soup. I've long since changed that, but there might be other edge cases like that remaining that might lead to some subtle bugs in the future.

ObjectStates (and SoupStates) are serializing and deserializing with object_id. The only thing I see now is that I did not edited but should be are eq and hash methods (I will push code soon with them changed). There is also repr that could be changed, but I'm not sure if it is desirable.

IMO adding object_id elsewhere is even more bug-prone so adding it to ObjectStates is the least bad way to track a single object.

bmielnicki:
> Overall looks good. I left some pretty nitpicky feedback but all the requested changes should be very minor, if anything at all. As a general comment, I would be very careful with the object_id field -- I like the idea of adding a unique id to each game object, but I can forsee a lot of pitfalls (i.e. a serialized and then deserialized object not keeping its uid) as well as logic pitfalls, as the rest of Overcooked_mdp.py was written without the uid in mind. There might be cases where we are creating new objects that would have previously been considered equal instead of keeping the old object around. For example, before we used to create a new Onion object when it was added to a soup. I've long since changed that, but there might be other edge cases like that remaining that might lead to some subtle bugs in the future.

ObjectStates (and SoupStates) are serializing and deserializing with object_id. The only thing I see now is that I did not edited but should be are eq and hash methods (I will push code soon with them changed). There is also repr that could be changed, but I'm not sure if it is desirable.

IMO adding object_id elsewhere is even more bug-prone so adding it to ObjectStates is the least bad way to track a single object.

bmielnicki:
btw. I've deleted one comment by mistake (I think it was about "events_list" field in episode info):
I think Trajectory class would help to clean up a bit code - then we need to store events in only one form (IMO storing full dictionaries as in events_list is the most flexible way to do it) and use object methods to generate information in other forms. It is very likely that most (if not all) of the extract_events.py will be part of the Trajectory class.

Many of the code is outdated anyway (i.e. not taking care of event adjectives, not distinguishing in visualization between item drop, potting, and soup delivery).
Most likely this code will be really useful after finishing https://github.com/HumanCompatibleAI/overcooked_ai/issues/46 when more visualizations will be added. We will also know then if we want to stick to "js runnable from python" paradigm or go into pure python solutions

testing/visualization_test.py
nathan-miller23: Ah gotcha, that's a cool way to test it. Testing against a specific list of variable names (like you do now) is much cleaner and clearer

nathan-miller23:
>Let me know if we want to change objects_ids to deterministic (e.g. ints counting from 0, use of seeded random functions) so we can use them in eq functions.

I definitely think we need to include the object_id in the `__eq__` method, at least that seems like the more intuitive way to say two objects are "equal". I would, however, include a utility function, like `uid_independent_equal` or something, that checks if two ObjectStates are equal in all areas except ID. This function can then be used for tests (sorta like how GridworldState has a `timestep_independent_equal` function). Another (perhaps cleaner) way to do it is to have the uid come from the numpy random suite and then set `np.random.seed` as you suggested. 

Both of these approaches seem equally viable, so I'll leave it to you to decide which is better.

src/overcooked_ai_py/visualization/visualization_utils.py
```diff
@@ -0,0 +1,55 @@
+import tempfile, webbrowser, json, uuid, os
+from IPython.display import display, HTML
+from string import Template
+from overcooked_ai_py.static import VISUALIZATION_DIR
+
+def load_visualization_file_as_str(filename):
+    with open(os.path.join(VISUALIZATION_DIR, filename), 'r') as f:
+        lines = f.readlines()
+        return "\n".join(lines)
+
```
bmielnicki, line 10: I don't like how this code (and its previous version based on pkgutil) is failing on osx test workflow despite other reading files work without problems. I will try to fix it later but if you have any idea why reading file can fail, but only on osx workflow let me know.

```
Traceback (most recent call last):
  File "testing/visualization_test.py", line 27, in test_create_chart_html
    html = create_chart_html(self.extracted_events1, box_id=box_id)
  File "/Users/runner/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/overcooked_ai_py/visualization/visualization_utils.py", line 48, in create_chart_html
    js_text_template = Template(load_visualization_file_as_str("event_chart.js"))
  File "/Users/runner/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/overcooked_ai_py/visualization/visualization_utils.py", line 7, in load_visualization_file_as_str
    with open(os.path.join(VISUALIZATION_DIR, filename), 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/overcooked_ai_py/visualization/event_chart.js'
```

testing/visualization_test.py
bmielnicki: done

src/overcooked_ai_py/visualization/extract_events.py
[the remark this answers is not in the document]
bmielnicki: done

bmielnicki:
> > Let me know if we want to change objects_ids to deterministic (e.g. ints counting from 0, use of seeded random functions) so we can use them in eq functions.
> 
> I definitely think we need to include the object_id in the `__eq__` method, at least that seems like the more intuitive way to say two objects are "equal". I would, however, include a utility function, like `uid_independent_equal` or something, that checks if two ObjectStates are equal in all areas except ID. This function can then be used for tests (sorta like how GridworldState has a `timestep_independent_equal` function). Another (perhaps cleaner) way to do it is to have the uid come from the numpy random suite and then set `np.random.seed` as you suggested.
> 
> Both of these approaches seem equally viable, so I'll leave it to you to decide which is better.

I've done it using method ids_independent_equal to not rely on identical order of calling RNG with the exact same seed at the start. There are some things that can be done to prevent randomness like setting random state in game state. After finishing coding I think that saving rng state approach would be probably a bit better but given that the current solution is already done and works I will leave it as it is.

src/overcooked_ai_py/visualization/visualization_utils.py
bmielnicki: solved, it was about -e flag in pip install

src/overcooked_ai_py/tutorial_notebooks/Agent_evaluator_intoduction.ipynb
```diff
@@ -0,0 +1,1387 @@
...
+    "- empty space - ' '\n",
+    "- counter - 'X'\n",
+    "- onion dispenser - 'O'\n",
+    "- tomato dispenser - 'T'\n",
+    "- pot (place where players cook soup from onions and tomatoes) - 'P' \n",
+    "- dish dispenser - 'D '\n",
+    "- serving location - 'S'\n",
+    "- player starting location - number  \n",
+    "  \n",
+    "You can save layout in ovecooked_ai/overcooked_ai_py/data/layouts directory and then run agent evaluator AgentEvaluator({\"layout_name\": layout_name}) where layout_name is filename without \".layoyut extension\".  \n",
```
bmielnicki, line 891: This notebook is not up to date with master branch. Probably needs some refinement before merging this file to master - see comment possible changes in  https://github.com/HumanCompatibleAI/overcooked_ai/issues/48 issue

bmielnicki:
Moving this PR to draft as soon I will change the chart code to add support of adjectives, delivery events, etc.
While I'm changing the visualization code I'm considering refactor to make code cleaner and more maintainable i.e. to make this and similar changes easier - ideally there is no need for modifying js code at all in case of adding a new adjective to events. Mostly it gonna be pushing all of the code distinguishing objects types/event adjectives/players etc into automatically created CSS classes, so d3.js would be used only for generation/modifying of HTML with well-named classes. A nice side effect of it would be allowing users to configure everything that is configurable by CSS.

bmielnicki:
I've added new functionalities and did a bit of refactoring. 

1. There is support for giving custom CSS code to create_chart_html function either as a CSS file path or CSS string. 

2. Adjectives are now visible on the chart. My idea for object adjectives was to use borders (as I did not have another idea) of a different colour.

3. Potting and deliveries have now separate visual representations (ellipse and checkmark).

4. There is now configurable what cumulative data is visible to the user (you now can choose event adjectives and event actions for each line in cumulative charts).
It is for parameter "cumulative_data_lines" with value looking like that:
`{"actions":["action1", "action2"], "adjectives":["adj1", "adj2"]}, "name": "Name of line in the legend",  "class": "css-class-name"}`

5. A previously big chunk of config that was about CSS properties inside python and js. If it was possible it was moved to the CSS file. If needed I can add a translation of the configuration in python dictionary directly to CSS code in case of most used stuff like chart width.
The biggest thing I did not move out of js code is the generation of SVG objects with data points (now pickups and drops are always triangles, potting is ellipse and delivery is checkmark and there is no way to check it without modifying js code).

6. The tutorial notebook is deleted as it would be added when making https://github.com/HumanCompatibleAI/overcooked_ai/issues/48 after the merge of native python visualization and this pull request.

Example chart below:
new-event-chart

bmielnicki:
I've split the event_chart.js file into 2 parts: render_events_chart function (used in both overcooked_demo and this PR) and script to rune render_events_chart from python (template with variables that are sent from python code).
Also, there is an added event that marks the end of the episode (so the x-axis is always the same length as the whole episode perfectly aligning slider handle of the same width).

micahcarroll:
Given that this is a lower priority issue right now, I'll temporarily close this PR for bookeeping – we can re-open it in the future if we are interested in exploring this direction further.

PR #40, opened 2020-08-07
Updated planners and agents for the new MDP + Housecleaning

mesutyang97:
Updated planners and agents for the new MDP

Discontinued support for Heuristic (and thus A* searches for coupled planning) for the new MDP. As a result, 
- CoupledPlanningAgent and EmbeddedPlanningAgent are deprecated
- parts of AgentEvaluator that uses CoupledPlanningAgent are deprecated


overcooked_ai_py/agents/agent.py
```diff
@@ -507,33 +422,24 @@ def ml_action(self, state):
...
             soup_nearly_ready = len(ready_soups) > 0 or len(cooking_soups) > 0
             other_has_dish = other_player.has_object() and other_player.get_object().name == 'dish'
 
             if soup_nearly_ready and not other_has_dish:
                 motion_goals = am.pickup_dish_actions(counter_objects)
             else:
-                next_order = None
-                if state.num_orders_remaining > 1:
-                    next_order = state.next_order
+                next_order = list(state.all_orders)[0]
```
micahcarroll, line 436: I think for all classes that only work for your version of the MDP it would be nice to have some assertion checks that make sure they're not being used on some of the unsupported MDPs. In particular in this instance, you could assert that len(state.all_orders) == 1 and that the only order is a 3-onion soup

overcooked_ai_py/agents/agent_test.py
```diff
@@ -66,316 +66,311 @@ def test_fixed_plan_agents(self):
...
+#         epa = EmbeddedPlanningAgent(other_agent, agent_evaluator.mlp, agent_evaluator.env, delivery_horizon=1)
+#         ap = AgentPair(epa, other_agent)
+#         agent_evaluator.evaluate_agent_pair(ap, num_games=1, display=DISPLAY)
+
+
+# Deprecated. Due to CoupledPlanningAgent
+# class TestScenarios(unittest.TestCase):
+#
+#     """
+#     Corridor: assuming optimality / planning horizon – scenario1_s
```
micahcarroll, line 149: I feel like these Scenario tests can be totally deleted, there is close to 0 chance that we will be revisiting this

overcooked_ai_py/agents/benchmarking.py
```diff
@@ -40,45 +40,46 @@ def __init__(self, mdp_params={}, env_params={}, mdp_fn=None, force_compute=Fals
...
-    def evaluate_one_optimal_one_greedy_human(self, num_games, h_idx=0, display=True):
-        h = GreedyHumanModel(self.mlp)
-        r = CoupledPlanningAgent(self.mlp)
-        agent_pair = AgentPair(h, r) if h_idx == 0 else AgentPair(r, h)
-        return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display)
-
-    def evaluate_agent_pair(self, agent_pair, num_games, game_length=None, start_state_fn=None, metadata_fn=None, metadata_info_fn=None, display=False, info=True):
+        return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display, regen_mdp=regen_mdp)
+
+    # Depreciated. Need to amend Heuristic before reactivating CoupledPlanningAgent:
```
micahcarroll, line 53: I believe this is a typo: "Deprecated". I think I might have seen this before elsewhere in the code, might be worth to command f in the codebase and replace all instances

overcooked_ai_py/mdp/mdp_utils.py
```diff
@@ -0,0 +1,537 @@
+import random
+import copy
+import numpy as np
+INTERACT_TRANSITION_COST = 2
+
+INFINITY = 999999999
```
micahcarroll, line 6: Using `np.inf` usually works well too

overcooked_ai_py/mdp/mdp_utils.py
```diff
@@ -0,0 +1,537 @@
+import random
+import copy
```
micahcarroll, line 2: Curious about where we use most of the utils defined in this file: is it for potentials and layout generation?

overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -76,21 +76,20 @@ def __init__(self, mdp_generator_fn, start_state_fn=None, horizon=MAX_HORIZON, m
         self.mlp_params = mlp_params
         self.start_state_fn = start_state_fn
         self.info_level = info_level
-        # self.lst_len = 8
-        # self.mdp_lst = [self.mdp_generator_fn() for _ in range(self.lst_len)]
+        self.lst_len = 4
+        self.mdp_lst = [self.mdp_generator_fn() for _ in range(self.lst_len)]
```
micahcarroll, line 80: Maybe I'm forgetting the conclusions about our previous conversations about this, so correct me if I'm wrong. I thought that we concluded that in most cases (eventually) we would have environments with infinitely many mdps (case which I don't think is currently supported).

Because of that, it seems to me like the cleanest way is not to have the `self.mdp_lst` attribute, and just have `self.mdp_generator_fn()` return one of 4 mdps at random (or even simply iterate between them deterministically). Is there any case in which we explicitly need to have `self.mdp_lst`?

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1,78 +1,409 @@
 import itertools, copy
```
micahcarroll, line 1: It's hard for me to tell which changes you made and which changes Nathan made in this file, so I haven't looked at it in depth. Let me know if there's anything you wanted me to look at more closely!

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1275,26 +1643,43 @@ def process_for_player(primary_agent_idx):
...
+                            if obj.is_ready:
+                                state_mask_dict["soup_done"] += make_layer(obj.position, 1)
                     else:
-                        raise ValueError("Unrecognized soup")
+                        # If player soup is not in a pot, treat it like a soup that is cooked with remaining time 0
+                        state_mask_dict["onions_in_soup"] += make_layer(obj.position, ingredients_dict["onion"])
+                        state_mask_dict["tomatoes_in_soup"] += make_layer(obj.position, ingredients_dict["tomato"])
+                        state_mask_dict["soup_done"] += make_layer(obj.position, 1)
+                    # else:
+                    #     raise ValueError("Unrecognized soup")
```
micahcarroll, line 1668:
```suggestion
```

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1247,10 +1603,19 @@ def process_for_player(primary_agent_idx):
...
             state_mask_dict = {k:np.zeros(self.shape) for k in LAYERS}
 
             # MAP LAYERS
+            if horizon - overcooked_state.timestep < 40:
+                # state_mask_dict["urgency"] = np.ones(self.shape) * (horizon - overcooked_state.timestep)
+                state_mask_dict["urgency"] = np.ones(self.shape)
+
+
+            # state_mask_dict["urgency"] = np.ones(self.shape) * (40 - min(40, horizon - overcooked_state.timestep))
+            # state_mask_dict["urgency"] = np.ones(self.shape) * 40
```
micahcarroll, line 1617: Just cleaning a couple things up, given that it's an important file, it's easy to miss these lines later, and these changes are easy to re-implement if necessary

```suggestion
```

overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1247,10 +1603,19 @@ def process_for_player(primary_agent_idx):
...
                         for i, d in itertools.product([primary_agent_idx, other_agent_idx], Direction.ALL_DIRECTIONS)]
 
-            LAYERS = ordered_player_features + base_map_features + variable_map_features
+            # LAYERS = ordered_player_features + base_map_features + variable_map_features
+            LAYERS = ordered_player_features + base_map_features + variable_map_features + urgency_features
             state_mask_dict = {k:np.zeros(self.shape) for k in LAYERS}
 
             # MAP LAYERS
+            if horizon - overcooked_state.timestep < 40:
+                # state_mask_dict["urgency"] = np.ones(self.shape) * (horizon - overcooked_state.timestep)
```
micahcarroll, line 1612:
```suggestion
```

overcooked_ai_py/planning/planners.py
```diff
@@ -1033,13 +1052,9 @@ def get_ml_plan(self, start_state, h_fn, delivery_horizon=4, debug=False):
...
         """
         start_state = start_state.deepcopy()
-        if start_state.order_list is None:
-            start_state.order_list = ["any"] * delivery_horizon
-        else:
-            start_state.order_list = start_state.order_list[:delivery_horizon]
 
         expand_fn = lambda state: self.get_successor_states(state)
-        goal_fn = lambda state: state.num_orders_remaining == 0
+        goal_fn = lambda state: state.delivery_rew >= DELIVERY_REW_THRES
```
micahcarroll, line 1057: Just to double check, we're currently not using any of this code right?

overcooked_ai_py/agents/agent.py
mesutyang97: Sounds good. Added the check and the error message if attempting to do more than 3-onion-soup order

overcooked_ai_py/agents/agent_test.py
mesutyang97: Removing these in the newest commit.

overcooked_ai_py/agents/benchmarking.py
mesutyang97: OK, sounds good.

overcooked_ai_py/mdp/mdp_utils.py
mesutyang97: updated

overcooked_ai_py/mdp/mdp_utils.py
mesutyang97: Functions in this file was basically a very stripped down version of planner (with the only API to find distance and plans, also ignoring orientations). But instead of pre-computing plans, it only compute plans on request.

This was first used in my version of the potential function. Nathan, in his latest PR for overcooked_ai, uses the full planner directly in the potential function.

I think it makes sense for him to use the full planner since he will be using a handful of layout only. But for the multi-layout, it is hard to come across the same layout multiple times, so I am not sure if we should use the full planner.

overcooked_ai_py/mdp/overcooked_env.py
mesutyang97: You are right. On the side, I was also running experiment on the effect of outer-shape on the side, and I usually first contrain the number of MDPs when getting a grip of where the hyperparameters should be starting (If I set it to be infinite, it will take many many iterations before I can know if the hyperparameters is even within the ball-park) But I will update those to reflect the infinite version like we agreed on before.

overcooked_ai_py/mdp/overcooked_mdp.py
mesutyang97: Nathan and I have kind of gone on different paths on this file. I will be spending a lot of time merging this file.

overcooked_ai_py/planning/planners.py
mesutyang97: This is right. I am deleting them

mesutyang97: I deprecated all the parts using the heuristic_fn (and thus the goal_fn)

overcooked_ai_py/mdp/mdp_utils.py
micahcarroll: Yeah that makes sense. I think we should add some text about where this is used at the top of the file. And if this is the only location where it's used, we might want to call this file `lightweight_planning.py` or `planning_ulils.py` something like this (feel free to choose a better name).

overcooked_ai_py/mdp/overcooked_env.py
micahcarroll: Yeah, mainly what i was saying was that even if you had a `self.mdp_generator_fn()` (that supports infinite # of MDPs), you could still represent the 1 (or few MDP case), by simply doing a variation of the following (in mostly pseudo-code):

```
mdps = [LayoutGenerator().generate_mdp(CUSTOM_SETTINGS) for _ in range(NUM_LAYOUTS)]
mdp_generator_fn()  = lambda: np.random.choice(mdps)
```

This should enable you to do testing of all kinds, if I understand correctly what you're trying to do.

PR #41, opened 2020-08-14
Recipes generation

bmielnicki:
Change includes automatic generation of recipes given additional parameters.
You can use recipes generation by supplying dicts into keys "generate_all_orders" and/or "generate_bonus_orders" in mdp_params.
Dict that configures generation process is currently (values are default values):
```
{
    "n": 1, # (int): how many recipes generate
    "min_size": 2, # (int): min generated recipe size
    "max_size": 3, # (int): max generated recipe size
    "ingredients": Recipes.ALL_INGREDIENTS, # (list(str)): list of ingredients used for generating recipes
    "recipes": Recipes.ALL_RECIPES # (list(Recipes) or list(dicts)) list of recipes (as dicts or Recipes) to choose from, by default it is Recipes.ALL_RECIPES for all_orders and all_orders for bonus_orders
}
```
i.e.
```
mdp_params = {"generate_all_orders": {"n":2, "ingredients": ["onion"], "min_size":2, "max_size":3},
"generate_bonus_orders": {"n":1, "min_size":2, "max_size":3}}
mdp_gen_params = {"mdp_params": mdp_params}  
mdp_fn = LayoutGenerator.mdp_gen_fn_from_dict(**mdp_gen_params)
env = OvercookedEnv(mdp_fn, **DEFAULT_ENV_PARAMS)
```

There is also also possibility to use feature types of choice for layout generation (to use tomato dispenser).


src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -76,29 +78,59 @@ def mdp_generator_fn():
...
+        """
+        mdp_params = copy.deepcopy(mdp_params)
+        if mdp_params.get("generate_all_orders"):
+            all_orders_kwargs = copy.deepcopy(mdp_params["generate_all_orders"])
+
+            if all_orders_kwargs.get("recipes"):
+                 all_orders_kwargs["recipes"] = [r if isinstance(r, Recipe) else Recipe.from_dict(r) for r in all_orders_kwargs["recipes"]]
+        
+            all_recipes = Recipe.generate_random_recipes(**all_orders_kwargs)
+            mdp_params["start_all_orders"] = [r.to_dict() for r in all_recipes]
```
nathan-miller23, line 100: It's important that `mdp_params` is JSON serializable, as saving and loading agents in the client code relies on this. Thus, you should probably only support recipe dicts as input, since Recipe instances are not JSON serializable.

bmielnicki: replaced `[r if isinstance(r, Recipe) else Recipe.from_dict(r) for r in all_orders_kwargs["recipes"]]` with `[Recipe.from_dict(r) for r in all_orders_kwargs["recipes"]]`

mesutyang97:
I think these are good features to have! Thanks for adding the functionalities and tests!

I guess there were some miscommunication, but I have been modified the layout_generator a month ago, and since Nathan doesn't need it in his project, I never pushed the changes (but it was recently staged along with other important changes to planner). 

The functionality you proposed will definitely be helpful in a long run. I will help resolve the merge conflicts.

Thanks again for the hard work!

src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -166,11 +198,11 @@ def dig_space_with_fringe_expansion(self, grid, prop_empty=0.1):
                 if grid.is_valid_dig_location(location):
                     fringe.add(location)
 
-    def add_features(self, grid, prop_features=0):
+    def add_features(self, grid, prop_features=0, feature_types=None):
         """
         Places one round of basic features and then adds random features 
         until prop_features of valid locations are filled"""
-        feature_types = [POT, ONION_DISPENSER, DISH_DISPENSER, SERVING_LOC] # NOTE: currently disabled TOMATO_DISPENSER
+        if not feature_types: feature_types = DEFAULT_FEATURE_TYPES
```
mesutyang97, line 205: @bmielnicki can we stremline the code to be `feature_types=DEFAULT_FEATURE_TYPES` in the argument stage?

bmielnicki: Main reason to do it this way is to prevent using mutables as default parameters. See "Mutable Default Arguments" section in https://docs.python-guide.org/writing/gotchas/ . 
It probably would not break the code now, but it is easy to break it by accident.
My bad that it does not work as intended, more safe (intended) version would be:
`if not feature_types: feature_types = copy.copy(DEFAULT_FEATURE_TYPES)`
Other approach is to use tuple as its immutable. I will change feature_types default parameter to the tuple.

mesutyang97:
Hey @bmielnicki . Sorry about the delay, but I have tried my best to merge your functionality into the updated master. There might be some minor syntax issues (because I am modifying signature in a few places to resolve the merge conflicts, but it should be pretty close. Let me know if you could kindly read the changes and let me know if the modification is performing as expected. Again, thanks a lot for the hardworl

bmielnicki:
> Hey @bmielnicki . Sorry about the delay, but I have tried my best to merge your functionality into the updated master. There might be some minor syntax issues (because I am modifying signature in a few places to resolve the merge conflicts, but it should be pretty close. Let me know if you could kindly read the changes and let me know if the modification is performing as expected. Again, thanks a lot for the hardwork

I've checked why tests failed and fixed what did not work in the test file (mostly correct format/content of LayoutGenerator.mdp_gen_fn_from_dict input) and in layout generator (biggest change is adding methods create_base_params and make_new_layout needed to generate new recipes when generating new layout).

mesutyang97:
Thanks a lot for address the issues! The changes look great! Trying to make the merge now

PR #49, opened 2020-08-15
Planner Update (Compatible with the new MDP) + Variable MDP Generator + Multi-Layout-Training Structure

mesutyang97:
Summary of changes:

Part 1: Planner changes
- Updated planner to be compatible with new MDP environment
- Updated test for planners (`planner_test.py`) and agent (`agent_test.py`)
- Deprecated inactive components that we will not be by the updated planner or ToM model (such as Heuristics, CoupledPlanningAgent, and a majority of medium level planner)

Part 2: Variable MDP Generator
- Layout Generator has been updated to generate MDP from a set of parameters
- mdp_fn is now the default way to generate variable MDP across different APIs
- Placed mlp (MediumLevelPlanner) inside OvercookedEnv, to enable planner reset under env reset

Part 3: Multi-layout training structure is enabled
- We can keep a finite list of Variable MDP in the same OvercookedEnv by using` LST_LEN` global flag
- when the flag is set to -1, we are not keeping a finite list, but instead doing completely infinite Variable MDP generation.

Part 4: MDP Variable Scheduling enabled
- now env.reset can take in outside_information when generation new `mdp_gen_params`


src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -12,74 +12,156 @@
...
+
+    def __init__(self, params_schedule_fn=None, mdp_params_always=None):
+        """
+        outer_shape: the outer shape of the mdp to be generated
+        schedule_fn: the function to produce a set of mdp_params for a specific layout
+        mdp_params_always: the mdp_params that the generator should always been using (no scheduling)
+        """
+        assert params_schedule_fn != None or mdp_params_always != None, \
+            "one of params_schedule_fn and mdp_params_always must be not Null"
+        self.mdp_params_always = mdp_params_always
```
micahcarroll, line 70: mdp_params_always can be considered a special case of params_schedule_fn where the schedule always returns the same params (similar to the mdp_generator_fn). Up to you whether you want to make this change – I don't have strong opinions about how the layout generator should be organized

src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -12,74 +12,156 @@
...
-            min_padding = (size_bounds[0][1], size_bounds[1][1])
-            layout_generator = LayoutGenerator(min_padding, mdp_params)
+
+            assert "inner_shape" in mdp_params.keys(), "mdp_params is missing inner_shape"
+            assert "prop_empty" in mdp_params.keys(), "mdp_params is missing prop_empty"
+            assert "prop_feats" in mdp_params.keys(), "mdp_params is missing prop_feats"
+            assert "display" in mdp_params.keys(), "mdp_params is missing display"
+            assert "start_all_orders" in mdp_params.keys(), "mdp_params is missing start_orders"
+            assert "recipe_values" in mdp_params.keys(), "mdp_params is missing recipe_values"
+            assert "recipe_times" in mdp_params.keys(), "mdp_params is missing recipe_times"
```
micahcarroll, line 142: You could replace this with something like
```suggestion
            required_keys = ["inner_shape", "prop_empty", "prop_feats", "display", "start_all_orders", "recipe_values", "recipe_times"]
            missing_keys = [k for k in required_keys if k not in mdp_params.keys()]
            assert len(missing_keys) == 0, "These keys were missing from the mdp_params: {}".format(missing_keys)
```

src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -12,74 +12,156 @@
...
+            inner_shape = mdp_params["inner_shape"]
```
micahcarroll, line 143: Might want to rename `mdp_params` to `mdp_gen_params` so as to avoid confusion with `mdp_params` in OvercookedGridworld

src/overcooked_ai_py/planning/planners.py
```diff
@@ -1,18 +1,22 @@
+# this denotes whether we are running the clean version of planners
+clean = True
```
micahcarroll, line 2: What does clean version of the planners mean? Elaborate in the comment

src/overcooked_ai_py/planning/planners.py
```diff
@@ -571,13 +581,17 @@ def _compute_plan_from_joint_graph(self, joint_start_state, joint_goal_state):
...
         """
         assert self.is_valid_joint_motion_pair(joint_start_state, joint_goal_state), joint_goal_state
         # Solve shortest-path graph problem
         start_positions = list(zip(*joint_start_state))[0]
         goal_positions = list(zip(*joint_goal_state))[0]
         joint_positions_node_path = self.joint_graph_problem.get_node_path(start_positions, goal_positions)[1:]
+        if len(joint_positions_node_path) == 0:
+            # The cost will be infinite if there is no path
+            num_player = len(goal_positions)
+            return [], None, [np.inf] * num_player
```
micahcarroll, line 594: Maybe do the `try`/`except` approach I was suggesting in the `search.py` (?) file

src/overcooked_ai_py/planning/planners.py
```diff
@@ -714,9 +729,7 @@ class MediumLevelActionManager(object):
     
     Args:
         mdp (OvercookedGridWorld): gridworld of interest
-        start_orientations (bool): whether the JointMotionPlanner should store plans for 
-                                   all starting positions & orientations or just for unique 
-                                   starting positions
+        params (dictionary): parameters for the medium level action manager
```
micahcarroll, line 732: Thanks for cleaning up some of the docstrings!! 🙏

src/overcooked_ai_py/planning/planners.py
```diff
@@ -762,6 +775,7 @@ def get_medium_level_actions(self, state, player, waiting_substitute=False):
         
         Args:
             state (OvercookedState): current state
+            player (PlayerState): the player's current state
```
micahcarroll, line 778: Again here too! 💯

testing/agent_test.py
```diff
@@ -66,316 +67,34 @@ def test_fixed_plan_agents(self):
...
+    #     a1 = CoupledPlanningAgent(self.mlp_large)
+    #     agent_pair = AgentPair(a0, a1)
+    #     start_state = OvercookedState(
+    #         [P((2, 2), n),
+    #          P((2, 1), n)],
+    #         {})
+    #     env = OvercookedEnv.from_mdp(large_mdp, start_state_fn=lambda: start_state)
+    #     trajectory, time_taken, _, _ = env.run_agents(agent_pair, include_final_state=True, display=DISPLAY)
+    #     end_state = trajectory[-1][0]
+
```
micahcarroll, line 82: We can probably remove deprecated tests too. My main philosophy for keeping commented out deprecated stuff is to indicate that "at one point, this functionality existed. If you want to actually resume the functionality, you should probably go back in the commit history and find everything related to this feature, copy it back into the code, and update it". That's why i usually recommend deleting minor things in the middle of other code / testing, but tend to try to keep commented out classes or class methods.

testing/planners_test.py
```diff
@@ -320,143 +339,336 @@ def check_joint_plan(self, joint_motion_planner, start, goal, times=None, min_t=
         if min_t is not None: self.assertEqual(len(action_plan), min_t)
         if times is not None: self.assertEqual(plan_lengths, times)
 
-
-class TestMediumLevelPlanner(unittest.TestCase):
-
+# Rewritten of Heuristic
```
micahcarroll, line 342: What does this comment mean?

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +21,75 @@ class AgentEvaluator(object):
...
     """
 
-    def __init__(self, mdp_params, env_params={}, mdp_fn_params=None, force_compute=False, mlp_params=NO_COUNTERS_PARAMS, debug=False):
+
+    def __init__(self, mdp_params={}, env_params={}, mdp_fn=None, force_compute=False, mlp_params=NO_COUNTERS_PARAMS, debug=False):
         """
         mdp_params (dict): params for creation of an OvercookedGridworld instance through the `from_layout_name` method
         env_params (dict): params for creation of an OvercookedEnv
-        mdp_fn_params (dict): params to setup random MDP generation
+        mdp_fn (callable function): a function that can be used to create mdp
```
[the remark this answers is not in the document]
mesutyang97: I went throught the docstrings for classes in the overcooked_env or overcooked_mdp classes and fixed what I can identify as out-of-date

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +21,75 @@ class AgentEvaluator(object):
...
-        return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display)
+    def evaluate_random_pair(self, num_games=1, all_actions=True, display=False, native_eval=False):
+        agent_pair = AgentPair(RandomAgent(all_actions=all_actions), RandomAgent(all_actions=all_actions))
+        return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display, native_eval=native_eval)
 
-    def evaluate_one_optimal_one_random(self, num_games, display=True):
-        a0 = CoupledPlanningAgent(self.mlp)
-        a1 = RandomAgent()
+    def evaluate_human_model_pair(self, num_games=1, display=False, native_eval=False):
+        a0 = GreedyHumanModel(self.env.mlp)
```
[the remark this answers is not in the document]
mesutyang97: I think removing MediumLevelPlanner should be another PR, because that would take quite some extra set of  structure work

src/overcooked_ai_py/data/layouts/cramped_room_o_1order.layout
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
mesutyang97: Nathan updated some of the old layout to have multiple orders (some of which have tomatos). So whenever there is a conflict, I put in the o_1order layout, meaning that layout only have 1 order of onion soup

src/overcooked_ai_py/agents/benchmarking.py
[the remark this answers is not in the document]
micahcarroll: Ok sounds good! Was just thinking we should get it in before having the pip distribution set up.

src/overcooked_ai_py/data/layouts/cramped_room_o_1order.layout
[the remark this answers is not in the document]
micahcarroll: Ah I see – I'm not quite sure by what you mean with "whenever there is a conflict" -> what was the issue in the case of `cramped_room.layout`? The original file still only has onion soups, but is in the old format (so could be updated to the new one, by copy pasting your current `cramped_room_o_1order.layout` into it)

src/overcooked_ai_py/agents/benchmarking.py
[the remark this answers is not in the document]
micahcarroll: Thanks so much Mesut! 🙏

src/overcooked_ai_py/data/layouts/unident.layout
```diff
@@ -4,9 +4,11 @@
                 X    P  1 X
```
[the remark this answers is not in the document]
mesutyang97: I just went through all the layouts in the folder, updated the few that I missed (because it wasn't used in the tests)

src/overcooked_ai_py/agents/benchmarking.py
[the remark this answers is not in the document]
mesutyang97: No worries, happy to help

src/overcooked_ai_py/agents/benchmarking.py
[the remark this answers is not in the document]
mesutyang97: Yes, agreed.

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +21,75 @@ class AgentEvaluator(object):
...
+    #     a0 = CoupledPlanningAgent(self.env.mlp)
+    #     a1 = RandomAgent()
+    #     agent_pair = AgentPair(a0, a1)
+    #     return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display)
+    #
+    # def evaluate_one_optimal_one_greedy_human(self, num_games, h_idx=0, display=True):
+    #     h = GreedyHumanModel(self.env.mlp)
+    #     r = CoupledPlanningAgent(self.env.mlp)
+    #     agent_pair = AgentPair(h, r) if h_idx == 0 else AgentPair(r, h)
+    #     return self.evaluate_agent_pair(agent_pair, num_games=num_games, display=display)
```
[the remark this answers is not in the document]
mesutyang97: removing in the next commit

src/overcooked_ai_py/data/layouts/cramped_room_o_1order.layout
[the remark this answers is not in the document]
mesutyang97: I do realize I messed that, sorry! Updated in the next commit!

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -145,22 +170,46 @@ def display_states(self, *states):
...
+            print("Timestep: {}\nJoint action taken: {} \t Reward: {} + shaping_factor * {}\nAction probs by index: {}\nState potential = {} \t Δ potential = {} \n{}\n".format(
+                    self.state.timestep,
+                    tuple(Action.ACTION_TO_CHAR[a] for a in a_t),
+                    r_t,
+                    env_info["shaped_r_by_agent"],
+                    action_probs,
+                    "",
+                    # self.mdp.potential_function(self.state, self.mlp.mp),
+                    "",
+                    # 0.99 * env_info["phi_s_prime"] - env_info["phi_s"], # Assuming gamma 0.99
```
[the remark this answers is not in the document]
mesutyang97: This is slightly more problematic. I moved this to the channel to ask for Nathan's confirmation. I think the problem here is that we don't have a default onion_value or tomato_value, since they are pretty layout-specific.  but without those information, we can't use the most up-to-date potential function. @nathan-miller23 Please chime in on this issue. Thanks!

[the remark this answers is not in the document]
mesutyang97: I already have a phi_s fix in the pipeline, but that won't work unless the potential function is fixed

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -145,22 +170,46 @@ def display_states(self, *states):
...
+                    env_info["shaped_r_by_agent"],
+                    action_probs,
+                    "",
+                    # self.mdp.potential_function(self.state, self.mlp.mp),
+                    "",
+                    # 0.99 * env_info["phi_s_prime"] - env_info["phi_s"],  # Assuming gamma 0.99
+                    self
+                ), file=f
+            )
+            f.close()
```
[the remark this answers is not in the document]
mesutyang97: Addressed. Thanks!

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -286,7 +353,11 @@ def run_agents(self, agent_pair, include_final_state=False, display=False, displ
         trajectory = []
         done = False
 
-        if display: print(self)
+        if display:
+            fname = 'results_client_temp/roll_out_' + str(time.time()) + '.txt'
+            f = open(fname, 'w+')
+            print(self, file=f)
+            f.close()
```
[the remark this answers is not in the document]
mesutyang97: I reformatted this, and placed the name of the folder under the evaluation_params, which is defaulted to None, and then the transition will be printed to the terminal

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -346,7 +418,8 @@ def get_rollouts(self, agent_pair, num_games, display=False, final_state=False,
             trajectories["env_params"].append(self.env_params)
             trajectories["metadatas"].append(metadata_fn(rollout_info))
 
-            self.reset()
+            # we do not need to regenerate MDP if we are getting a series of rollout using the same mdp
```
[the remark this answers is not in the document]
mesutyang97: I added a bit more explanation in the next commit

src/overcooked_ai_py/planning/search.py
```diff
@@ -217,7 +217,22 @@ def _get_node_index_path(self, start_index, goal_index):
...
+            # X     2 P   --->  X     1 P
+            # X X X X X         X X X X X
+            # This is actually an absolutely impossible transition
+
+            # Override the error for now and return an empty list
+            # raise NotConnectedError("No path could be found from {} to {}".format(self._decoder[start_index], self._decoder[goal_index])
+            #                        + "This could be caused by using another layout's planner on this layout")
+
+            # here, we are returning a list with just the start_index, so that
+            # after taking the plan, the upstream will be left with nothing, enabling detection of this edge case
```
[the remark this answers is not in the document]
mesutyang97: Addressed

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -68,17 +73,36 @@ def __init__(self, mdp_generator_fn, start_state_fn=None, horizon=MAX_HORIZON, i
...
                                             "instance, use the OvercookedEnv.from_mdp method"
         self.variable_mdp = _variable_mdp
-        self.mdp_generator_fn = mdp_generator_fn
+        # infinite case
+        if LST_LEN == -1:
+            self.mdp_generator_fn = mdp_generator_fn
+        # finite MDP LST case
+        else:
+            self.mdp_lst = [mdp_generator_fn() for _ in range(LST_LEN)]
+            self.mdp_generator_fn = lambda : random.choice(self.mdp_lst)
```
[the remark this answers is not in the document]
mesutyang97: I have a few questions about this approach:
1. how do we know how many mdps we are using before we start training?
2. the original design was designed with the option of curriculumn in mind (the mdp_generator_fn could carry curriculumn information in it, so the long term plan was to replace env.mdp_lst after the agents were doing well enough in the current layouts

src/overcooked_ai_py/mdp/layout_generator.py
mesutyang97: I left this in to be backwards compatible (because some part of the old code calls `mdp_gen_fn_from_dict` using the mdp_params dictionary), especially if we want to use it to generate mdp for evaluators

src/overcooked_ai_py/mdp/layout_generator.py
mesutyang97: Addressed

src/overcooked_ai_py/planning/planners.py
mesutyang97: This was added because of circular import of OvercookedEnv. It has been addressed. Thanks for catching it!

src/overcooked_ai_py/planning/planners.py
mesutyang97: Addressed! Thanks for the wonderful suggestion!

testing/agent_test.py
mesutyang97: Sounds good

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -346,7 +406,10 @@ def get_rollouts(self, agent_pair, num_games, display=False, final_state=False,
             trajectories["env_params"].append(self.env_params)
             trajectories["metadatas"].append(metadata_fn(rollout_info))
 
-            self.reset()
+            # we do not need to regenerate MDP if we are trying to generate a series of rollouts using the same MDP
+            # Basically, the FALSE here means that we are using the same layout and starting positions
+            # (if regen_mdp == True, resetting will call mdp_gen_fn to generate another layout & starting position)
+            self.reset(regen_mdp=False)
             agent_pair.reset()
```
mesutyang97, line 413: New comments here to explain the flag

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -275,18 +328,24 @@ def execute_plan(self, start_state, joint_action_plan, display=False):
...
-
-        if display: print(self)
+        # default is to not print to file
+        fname = None
+
+        if dir != None:
+            fname = dir + '/roll_out_' + str(time.time()) + '.txt'
+            f = open(fname, 'w+')
+            print(self, file=f)
+            f.close()
```
mesutyang97, line 348: Now dir is passed down as an argument from evaluation_params. @micahcarroll  please let me know how this solution looks to you.

testing/planners_test.py
```diff
@@ -320,143 +339,336 @@ def check_joint_plan(self, joint_motion_planner, start, goal, times=None, min_t=
         if min_t is not None: self.assertEqual(len(action_plan), min_t)
         if times is not None: self.assertEqual(plan_lengths, times)
 
-
-class TestMediumLevelPlanner(unittest.TestCase):
-
+# Rewritten because the previous test depended on Heuristic, and Heuristic has been deprecated
+class TestMediumLevelPlannerSimple(unittest.TestCase):
```
mesutyang97, line 343: @micahcarroll re-written here

testing/planners_test.py
mesutyang97: clarified

src/overcooked_ai_py/data/layouts/cramped_room_o_1order.layout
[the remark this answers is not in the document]
micahcarroll: Yeah no worries!

src/overcooked_ai_py/data/layouts/unident.layout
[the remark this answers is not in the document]
micahcarroll: Ok sounds good, that should do for now!

src/overcooked_ai_py/mdp/overcooked_env.py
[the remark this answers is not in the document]
micahcarroll: I see, sounds good!

src/overcooked_ai_py/mdp/layout_generator.py
micahcarroll: We don't actually care about backwards compatibility for layoutgenerator code as it was never really used anywhere

src/overcooked_ai_py/mdp/overcooked_env.py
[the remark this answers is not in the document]
micahcarroll: I'm not quite sure I understand your first question: don't you always know how many MDPs you are training on before starting training? To answer your question literally (don't think this is what you mean), you can know the number of layouts you're training on by just looking at the length of the mdp_list (which you know to be infinity if mdp_lst is none). 

Generally here I'm operating under the assumption that the Env is not changing during training: the mdp_generator_fn might be changing subtly over the course of training, but it will not be changing from like a single layout training to few-layout training, to infinite-layout training. In general, it's almost always a bad to train on few layouts at the beginning if eventually you're intending to train on infinitely many, because you will overfit.

For the changing of the generator fn during the course of training, we should figure out what APIs make sense to do that – you could either keep track of how many times the gen function was called within the Generator logic (in order to advance in the schedule), or maybe passing in some info to the generator function, rather than having it have no arguments).

In any case, I'm not sure why your curriculum should have any relationship to mdp_lst whatsoever?

src/overcooked_ai_py/mdp/layout_generator.py
mesutyang97: I am sorry, I think I misread your initial comment, and made an ambiguous response because of my understanding. I will addressed this in the newest commit.

src/overcooked_ai_py/mdp/overcooked_env.py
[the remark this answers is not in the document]
mesutyang97: I see what you are saying. I think we can talk more about this later, but based on my observation of ppo learning on a finite set of layouts, the algorithm learns one layout at a time (basically, it will start with getting 0 on all the layouts. And then for 1 layout it will gradually learn to get up to ~100 sparse reward while all other layouts are still 0. And then it will start to conquer another layout, getting ~ 100 sparse reward, while the not-yet-learnt set of layout still all get 0). That's why I believe it might be good to dynamically change the size of mdp_lst in the future.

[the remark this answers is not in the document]
mesutyang97: But, I have made the change per your request, and it is on the newest commit

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -163,16 +165,22 @@ def print_state_transition(self, a_t, r_t, env_info, fname=None):
...
                     tuple(Action.ACTION_TO_CHAR[a] for a in a_t),
                     r_t,
                     env_info["shaped_r_by_agent"],
                     action_probs,
-                    "",
-                    # self.mdp.potential_function(self.state, self.mlp.mp),
-                    "",
-                    # 0.99 * env_info["phi_s_prime"] - env_info["phi_s"], # Assuming gamma 0.99
+                    state_potential_str,
+                    potential_diff_str,
```
mesutyang97: The output formulation has been streamlined

src/overcooked_ai_py/mdp/overcooked_env.py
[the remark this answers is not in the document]
mesutyang97: After talking to Nathan, this has been fixed in the newest commit

mesutyang97:
I have deprecated the `MediumLevelPlanner`, and we will only use `MediumLevelActionManager` moving forward. Details for deprecation can be found at commit cf3edec03a1171278af67f2bfddf7cf1b929a91b

src/overcooked_ai_py/agents/benchmarking.py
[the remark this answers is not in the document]
mesutyang97: I have completed deprecating MLP. Mark as closed.

src/overcooked_ai_py/mdp/layout_generator.py
micahcarroll: Ah no worries! :)

src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -67,17 +67,18 @@ def __init__(self, params_schedule_fn=None, mdp_params_always=None):
         """
         assert params_schedule_fn != None or mdp_params_always != None, \
             "one of params_schedule_fn and mdp_params_always must be not Null"
-        self.mdp_params_always = mdp_params_always
-        self.params_schedule_fn = params_schedule_fn
+        if mdp_params_always != None:
```
micahcarroll, line 70: We don’t have to incorporate this into this review specifically (and in fact this is up to you), but what I was suggesting was to remove mdp_params_always entirely from the `__init__` signature: if you want to keep the ability to easily specify a params dict that stays constant over time, you could have a `classmethod` `.from_fixed_params(fixed_params)`, similar to `from_mdp(mdp)` for OvercookedEnv. I usually tend to like having the `__init__` logic and signatures as simple as possible, keeping the `__init__` signature have the most general case, and if necessary adding special-case logic through classmethods.

mesutyang97: I see what you mean. Sounds reasonable. This has been updated in the newest commit

src/overcooked_ai_py/mdp/layout_generator.py
```diff
@@ -59,19 +59,18 @@ def DEFAILT_PARAMS_SCHEDULE_FN(outside_information):
...
-        else:
-            self.params_schedule_fn = params_schedule_fn
+        assert callable(params_schedule_fn), "params scheduling function must be a callable"
+        self.params_schedule_fn = params_schedule_fn
+
+    @staticmethod
+    def from_fixed_param(mdp_params_always):
+        # s naive schedule function that always return the same set of parameter
+        naive_schedule_fn = lambda x: mdp_params_always
+        return MDPParamsGenerator(naive_schedule_fn)
```
mesutyang97: simplified MDAPParamsGenerator

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +19,65 @@ class AgentEvaluator(object):
...
+            # infinite mdp
+            if 'num_mdp' not in env_params or env_params['num_mdp'] == -1:
+                self.mdp_fn = mdp_fn
+                self.env = OvercookedEnv(self.mdp_fn, **env_params)
+            else:
+                num_mdp = env_params['num_mdp']
+                assert num_mdp > 0, "invalid number of mdp"
+                self.mdp_lst = [mdp_fn() for _ in range(num_mdp)]
+                self.mdp_fn = lambda : np.random.choice(self.mdp_lst)
+                self.env = OvercookedEnv(self.mdp_fn, **env_params)
```
micahcarroll, line 49: Regarding the if/else logic in the init, we might want to have this handled with `@classmethod`s here too. Given that OvercookedEnv now works based off of `mdp_generating_fn`, we might even be able to deprecate the `mdp_fn` `__init__` signature parameter, and clean things up here more.

For example, one downside of this approach is that having `self.mdp_lst = [mdp_fn() for _ in range(num_mdp)]` is making strong assumptions on how the mdp_fn() will return different mdps sequentially at each call, and the following line is forcing a specific sampling function (which i guess in the case of the AgentEvaluator would probably usually make sense, but it would be nice to have this logic outside of the class or in a classmethod rather than in conditional logic in an init so that it's flexible to customize for strage use-cases).

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +19,65 @@ class AgentEvaluator(object):
...
+        if native_eval:
+            return self.env.get_rollouts(agent_pair, num_games=num_games, display=display, dir=dir, info=info,
+                                         metadata_fn=metadata_fn, metadata_info_fn=metadata_info_fn)
+        else:
+            horizon_env = self.env.copy()
+            horizon_env.horizon = self.env.horizon if game_length is None else game_length
+            horizon_env.start_state_fn = self.env.start_state_fn if start_state_fn is None else start_state_fn
+            horizon_env.reset()
+            return horizon_env.get_rollouts(agent_pair, num_games=num_games, display=display, dir=dir, info=info,
+                                            metadata_fn=metadata_fn, metadata_info_fn=metadata_info_fn)
```
micahcarroll, line 80: I'm not entirely sure in what cases doing the additional steps of copying the environment and resetting it should cause things to break for multi-layout evaluation? In fact, I feel like we probably want to build the multi-layout environment to work around this constraint, that seems quite intuitive: "If i have an environment, I copy it and reset it, it should still be the same environment." To add support for the case of environments with multiple MDPs, I'm guessing some additional optional params might need to be added to the `reset` function – in fact, you have already added `regen_mdp` (or something along those lines), that seems like a good idea. In fact, if `regen_mdp=False`, I'm guessing that this function should work as you intended (that might just require adding `regen_mdp` to the signature of `evaluate_agent_pair`)? Not sure what exact use case you had in mind?

Another thing that comes to mind as a useful signature addition to `.reset()` is `schedule_loc` (a 0 to 1 float indicating how far in we are in the schedule, that is also passed into the mdp generation fn) – as I was alluding to in some other comments.

In any case, this code could be simplified to the following, in order to have less repeats:
```suggestion
    def evaluate_agent_pair(self, agent_pair, num_games, game_length=None, start_state_fn=None, metadata_fn=None, metadata_info_fn=None, display=False, dir=None, info=True, native_eval=False):
        # this index has to be 0 because the Agent_Evaluator only has 1 env initiated
        # if you would like to evaluate on a different env using rllib, please modifiy
        # rllib/ -> rllib.py -> get_rllib_eval_function -> _evaluate

        # native eval: using self.env in evaluation instead of creating a copy
        # this is particulally helpful with variable MDP, where we want to make sure
        # the mdp used in evaluation is the same as the native self.env.mdp
        if not native_eval:
            horizon_env = self.env.copy()
            horizon_env.horizon = self.env.horizon if game_length is None else game_length
            horizon_env.start_state_fn = self.env.start_state_fn if start_state_fn is None else start_state_fn
            horizon_env.reset()
        return self.env.get_rollouts(agent_pair, num_games=num_games, display=display, dir=dir, info=info,
                                         metadata_fn=metadata_fn, metadata_info_fn=metadata_info_fn)
```

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -20,83 +19,65 @@ class AgentEvaluator(object):
...
+        force_compute (bool): whether should re-compute MediumLevelActionManager although matching file is found
+        mlam_params (dict): the parameters for mlam, the MediumLevelActionManager
+        debug (bool): whether to display debugging information on init
         """
         assert type(mdp_params) is dict, "mdp_params must be a dictionary"
-
-        if mdp_fn_params is None:
+        assert (mdp_params != {} and env_params != {}) or mdp_fn != None, "either evaluate from params or fn"
+        env_params["mlam_params"] = mlam_params
+        if mdp_fn is None:
```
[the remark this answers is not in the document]
mesutyang97: added two static function (`from_mdp` and `from_mdp_params`) to address these issues.

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: After exploring many option, I think I have one method figured out, and I am adding tests to make sure it is robust. Will update in the next commit

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: directly copying and regenerating will be a problem, since "copying" the OvercookedEnv only copies the generator function, not the mdp inside it. Copy the mdp inside it won't be good enough, because the reset in the env is going to generate a new mdp anyway. (Basically, copying over the lottery box (env) doesn't mean that the same guy will end up winning the lottery (env.mdp).)

Regarding passing in additional parameter: I have left space in the code to be able to do this. But I feel like it really should be in a different pull request, where we have a better idea about how we want to do MDP parameter generation.

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -19,36 +20,76 @@ class AgentEvaluator(object):
...
+        for mdp in mdp_lst:
+            assert type(mdp) == OvercookedGridworld, "some mdp in the mdp_lst are not OvercookedGridworld"
+
+        if sampling_freq is None:
+            mdp_fn = lambda : np.random.choice(mdp_lst)
+        else:
+            assert len(sampling_freq) == len(mdp_lst), "sampling frequency is not the same length as the mdp_lst"
+            assert np.sum(sampling_freq) == 1.0
+            mdp_fn = lambda : np.random.choice(mdp_lst, p=sampling_freq)
+        return AgentEvaluator(env_params, mdp_fn, force_compute, mlam_params, debug)
```
mesutyang97: Three staticmethod has been added, the `from_mdp_lst` gives the the possibility to do any arbitrary sampling frequency

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -19,36 +20,76 @@ class AgentEvaluator(object):
...
-                assert num_mdp > 0, "invalid number of mdp"
-                self.mdp_lst = [mdp_fn() for _ in range(num_mdp)]
-                self.mdp_fn = lambda : np.random.choice(self.mdp_lst)
-                self.env = OvercookedEnv(self.mdp_fn, **env_params)
+                assert type(num_mdp) == int and num_mdp > 0, "invalid number of mdp: " + str(num_mdp)
+                mdp_lst = [mdp_fn_naive() for _ in range(num_mdp)]
+                mdp_fn = lambda : np.random.choice(mdp_lst)
+        if "outer_shape" in env_params:
+            del env_params["outer_shape"]
+        return AgentEvaluator(env_params, mdp_fn, force_compute, mlam_params, debug)
```
micahcarroll: I'm really sorry for not being clearer about this, but in general I feel like it's good practice to avoid a lot of conditional logic within methods (i.e. having one method do 3 different tasks), and instead have one method for each if/else condition (i.e. for each "special case"). In the `__init__` method this is particularly important, but I feel like this also applies to all other methods too. 

In this case, you can have one method to generate from a layout name, one for infinite variable mdp, and one for finite variable mdp (using a layout generator). Usually, there is a nice way to reuse some of the methods for others, creating some heirarchy of abstraction that enables you to not have to re-write the same lines 3 times.

In particular, looking at the other methods below, it seems like a method for single mdps already exists (`from_mdp`). If you want another method to create a AgentEvaluator from layout name, you could create a `from_layout_name` classmethod that calls `from_mdp` once you have created the MDP as `OvercookedGridworld.from_layout_name` (to avoid having code duplicates). And for the case of multiple MDPs created through layout_generator, you might want to create a method that is named more specifically invoking that purpose. You could probably also collapse the finite case into the infinite case, and remove another conditional statement, or if that turns out to be too hard, you can make them into two different classmethods.

src/overcooked_ai_py/agents/benchmarking.py
```diff
@@ -19,36 +20,76 @@ class AgentEvaluator(object):
...
+            mdp_fn = lambda: mdp
+        else: # general case: variable mdp
+            assert outer_shape is not None, "outer_shape needs to be defined for variable mdp"
+            mdp_fn_naive = LayoutGenerator.mdp_gen_fn_from_dict(mdp_params, outer_shape, mdp_params_schedule_fn)
             # infinite mdp
-            if 'num_mdp' not in env_params or env_params['num_mdp'] == -1:
-                self.mdp_fn = mdp_fn
-                self.env = OvercookedEnv(self.mdp_fn, **env_params)
-            else:
+            if 'num_mdp' not in env_params or np.isinf(env_params['num_mdp']):
```
micahcarroll, line 54: We should probably operate under the assumption that num_mdp is always in env_params – it's more in line with offensive programming practices (that I feel like are better in research-code, as we want to catch errors or abstraction violations as soon as possible, and don't care about having code that errors when something is wrong – errors are not costly, as they don't impact any users or anything).

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: I agree with what you said, and I think it makes sense.

The current version of __init__ is able to accommodate the need you describe (crazy sampling scheme) by taking in the `mdp_fn` only. Could you let me know if this was not the desired behavior?

The staticmethod was introduced to save work if we don't want to hand write custom sampling scheme. Instead, we just pass in a mdp_lst and a sampling frequency and have the staticmethod do the dirty job. That was why I included the possibility to do a custom sampling frequency.

I will give the Clean Coding series a lot.

testing/common_test_generator.py
```diff
@@ -11,7 +11,7 @@ def generate_serialized_trajectory(mdp, save_path):
     sparse_reward = 0
```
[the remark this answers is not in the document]
mesutyang97: I can verify this is still being used in overcoooked_test, so deleting it will result in case failure. Nathan put those tests together so maybe he should answer this questions..

src/overcooked_ai_py/agents/benchmarking.py
micahcarroll: Yeah that makes sense – I guess I was wondering whether you actually expect to use a custom sampling scheme, rather than just sample uniformly. If not, we could remove the custom `sampling_freq`, and if somebody really wants to do that, they can just use the `__init__` method using `mdp_fn`. Alternatively you could keep the `sampling_freq` param, and clean the code this way

```
if sampling_freq is None:
   sampling_freq = np.ones(len(mdp_lst)) / len(mdp_lst)

mdp_fn = lambda : np.random.choice(mdp_lst, p=sampling_freq)
```

micahcarroll: Another code improvement here is:

```
assert all(type(mdp) == OvercookedGridworld for mdp in mdp_lst), "some mdp in the mdp_lst are not
```

testing/common_test_generator.py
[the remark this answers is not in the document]
micahcarroll: @nathan-miller23 do you know if we still use this file / where? My understanding is that we javascript doesn't even have encodings / dynamics anymore on their side, so I think it can be deleted – but wanted to double check with you

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: I see, thanks for the clarification!! This will be addressed in the newest commit.

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: Addressed, marking as resolved

src/overcooked_ai_py/agents/benchmarking.py
mesutyang97: This scheduling functionality and tests has been added on the newest commit b16c69b8b389cb38342ac70d93ebbdfe5368520a

Regarding the evaluate_agent_pair, I don't think the two are necessarily equivalent, because one is running from `self.env` and the other one is running on the cloned `horizon_env`. We need to have two different return statements, otherwise it will not produce the desired behavior.

mesutyang97:
Closed all the sections that have been addressed. 

Updated static initialization to get rid of if/else in init. 

Added scheduling functionality to environment reset.

Please advise. Thanks for all the help!!

src/overcooked_ai_py/agents/benchmarking.py
micahcarroll: horizon_env is usually equivalent to self.env (unless horizon or start state parameters are overwritten). But yeah, I agree that for now we can keep it this way!

mesutyang97: Sounds good

testing/schedule_test.py
```diff
@@ -0,0 +1,171 @@
+import unittest
```
micahcarroll, line 1: I suggest renaming this to `mdp_gen_schedule_test.py`

mesutyang97: Addressed! Thanks a lot!

Issue #47, opened 2020-08-14
Code Coverage Monitor


nathan-miller23:
Incorporating code coverage automation reports into the project to better enhancing testing and confidence in code. Example of such a tool [here]( https://codecov.io/)

nathan-miller23:
The `upload coverage to Codecov` portion of the integration test failed when the checks run in the current master branch. I assume this is some sort of connection issue connecting to the codecov client? The actual tests appeared to pass. https://github.com/HumanCompatibleAI/overcooked_ai/runs/995100873

bmielnicki:
It can happen if there is an issue with uploading the result to codecov. I've clicked "rerun all jobs" button and now it passed without errors so it is not an issue with the code itself. Possible solutions:
 - change flag fail_ci_if_error to false, but then we are not notified if there is fail in sending coverage data into the cloud
- add separate test GitHub workflow for coverage tests 
	- needs to run all the tests one more time
	- it is seen from pull request GUI which workflow failed (and if it is only coverage one or test one too)

micahcarroll:
Setting `fail_ci_if_error` to False seems fine to me. As long as it succeeds some of the time to upload it, I think we should be good.

PR #51, opened 2020-08-18
Update pythontests.yml

bmielnicki:
set fail_ci_if_error flag to false


micahcarroll:
( Was finishing dealing with #47 )

PR #55, opened 2020-09-06
harl merge fix

mesutyang97:
Minor fixes to ensure that this module will function correctly, when used by `human_aware_rl`

1. configure recipe if necessary when generating layout
2. default motion planner (`mp`) initialization after mdp generation. This was a hard one to identify and fix. The issue during the harl merge update is that when mlam_params contain `counter_goals` that is not empty, it will crash the `MotionPlanner` when we load saved `mp` from pickles. A lot of approches have been tried to fix this. The other way to address it would be to add default ingradients (like ('onion', 'onion', 'onion')) to the `__new__` method of the `Recipe` class. That method was outlined in 25b848eb47a6b7dd50f75d10bf3608decb8fe9a9 and that is a standalone fix as well. Happy to proceed with either way.


nathan-miller23:
I am confused as to why a lot of the MotionPlanner updates that are currently in the master branch are not in this PR. For example, the MotionPlanner class has staticmethods like `from_pickle_or_compute` that are missing

mesutyang97:
The MotionPlanner updates are all there, including `from_pickle_or_comput` (see the `planners.py` on `harl_merge_fix` [here](https://github.com/HumanCompatibleAI/overcooked_ai/blob/harl_merge_fix/src/overcooked_ai_py/planning/planners.py))

This update pulled from the most updated master (the 1.0.4 release), and only made very small adjustments.

You should be able to see that all the diffs [here](https://github.com/HumanCompatibleAI/overcooked_ai/pull/55/files)

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -252,7 +252,8 @@ def reset(self, regen_mdp=True, outside_info={}):
         if regen_mdp:
             self.mdp = self.mdp_generator_fn(outside_info)
             self._mlam = None
-            self._mp = None
+            # The motion planner (self.mp) will almost certainly be used. Initialize here to avoid pickle loading issues
+            self.mp
```
nathan-miller23, line 256: Do you not still need the `self._mp = None` line here in order to cause the `self.mp` line to recompute the motion planner? My understanding here is that it'll still return the old one since it was never invalidated

nathan-miller23: Also this might be too complicated for it to be worth it, but would it be possible to have `self._mp` point at `self._mlam.motion_planner.motion_planner`? The way it's set up now I believe we are redundantly computing two separate copies of the same `MotionPlanner`

nathan-miller23 (commented):
Looks good! Luckily the updates needed here were pretty minimal. After this branch and the human_aware_rl branch get synched we should update the PyPi release version for this repo

src/overcooked_ai_py/mdp/overcooked_env.py
mesutyang97: Thanks for pointing out the issue! It has been addressed by restoring `self._mp =None`, but upon calling `self.mp`, the `env` will first look into `self.mlam.motion_planner`

micahcarroll (approved):
LGTM 👍

PR #64, opened 2020-09-27
fix GreedyHumanModel so it cooks soups now

bmielnicki:
GreedyHumanModel could not cook any soups before. Now it can cook soups.

An additional improvement would be to cook any soup of size bigger than next_order too, or soups with ingredients that do not match the next order (e.g soups that has a tomato when the only order is ["onion", "onion", "onion"] to free up the pots. This improvement affects mostly the pairing of GreedyHumanModel with irrational agents acting at random.


mesutyang97 (approved):
We already discussed this. Looks good to me and thanks!

micahcarroll:
One thing we should do is add an assertion errors (and add to the docstring) that this GreedyHumanModel is only to be used with 3-onion soups.

Changing the GreedyHumanModel caused the featurization tests to fail because they actually [depend on](https://github.com/HumanCompatibleAI/overcooked_ai/blob/ade5cebde8cc9f8794c663edec530474d31e9402/testing/overcooked_test.py#L729) the GreedyHumanModel for generating them. We should probably also rename that `rnd_agent_pair` variable to `greedy_human_model_pair` (the naming is quite terrible).

micahcarroll (changes requested):
As described above

bmielnicki:
> One thing we should do is add an assertion errors (and add to the docstring) that this GreedyHumanModel is only to be used with 3-onion soups.
> 
> Changing the GreedyHumanModel caused the featurization tests to fail because they actually [depend on](https://github.com/HumanCompatibleAI/overcooked_ai/blob/ade5cebde8cc9f8794c663edec530474d31e9402/testing/overcooked_test.py#L729) the GreedyHumanModel for generating them. We should probably also rename that `rnd_agent_pair` variable to `greedy_human_model_pair` (the naming is quite terrible).

Done. I've also commented out the tomato-related motion goals.

micahcarroll (approved):
LGTM 👍

PR #67, opened 2020-10-06
Eval visual merge

mesutyang97:
Issues addressed with this PR:

1. bug in layout generator due to numpy usage
2. default information behavior: it is now being stored in case nothing is fed in.


src/overcooked_ai_py/agents/agent.py
```diff
@@ -319,7 +319,7 @@ def action(self, state):
 
             # NOTE: Assumes that calls to the action method are sequential
             self.prev_state = state
-        return chosen_action, {"action_probs": action_probs}
+        return chosen_action, {"action_probs": [action_probs]}
```
bmielnicki, line 322: I don't like incinsistency between RandomAgent and GreedyHumanModel action method output.
Example code:
```
from overcooked_ai_py.agents.agent import GreedyHumanModel, RandomAgent
from overcooked_ai_py.mdp.layout_generator import LayoutGenerator
from overcooked_ai_py.agents.benchmarking import AgentEvaluator

mdp_gen_params = {"layout_name": 'cramped_room'}
mdp_fn = LayoutGenerator.mdp_gen_fn_from_dict(mdp_gen_params)
env_params = {"horizon": 10}
agent_eval = AgentEvaluator(env_params=env_params, mdp_fn=mdp_fn)

trajectory = agent_eval.evaluate_human_model_pair(num_games=1)
state = trajectory["ep_states"][0][5]
grid = trajectory["mdp_params"][0]["terrain"]
greedy = GreedyHumanModel(mlam=agent_eval.env.mlam)
random = RandomAgent()
greedy.set_agent_index(0)
greedy_probs = greedy.action(state)[1]["action_probs"]
random_probs = random.action(state)[1]["action_probs"]
print("greedy agent probs:", greedy_probs, type(greedy_probs), "random_agent_probs:", random_probs, type(random_probs))
```
Output:
```
greedy agent probs: [array([0., 0., 1., 0., 0., 0.])] <class 'list'> random_agent_probs: [0.2 0.2 0.2 0.2 0.2 0. ] <class 'numpy.ndarray'>
```

mesutyang97: That is strange. I had the opposite problem earlier, but I will revert it back, now that I see this,

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -215,7 +215,10 @@ def step(self, joint_action, joint_agent_action_info=None, display_phi=False):
         """
         assert not self.is_done()
         if joint_agent_action_info is None: joint_agent_action_info = [{}, {}]
-        next_state, mdp_infos = self.mdp.get_state_transition(self.state, joint_action, display_phi, self.mp)
+        if not display_phi:
+            next_state, mdp_infos = self.mdp.get_state_transition(self.state, joint_action)
+        else:
+            next_state, mdp_infos = self.mdp.get_state_transition(self.state, joint_action, display_phi, self.mp)
```
mesutyang97, line 221: @nathan-miller23 let me know if this change will be OK for you. I put in this optimization so that it is not creating mp when not necessary

bmielnicki: Is there a reason why you wrote here `if not display_phi:` instead of `if display_phi:`? Latter seems more natural:
```
        if display_phi:
           next_state, mdp_infos = self.mdp.get_state_transition(self.state, joint_action, display_phi, self.mp)
        else:
            next_state, mdp_infos = self.mdp.get_state_transition(self.state, joint_action)
```
Anyway, it's not changing functionality and the current version is readable too so its not a big issue. I'm rather asking from curiosity.

bmielnicki (approved):
LGTM

src/overcooked_ai_py/mdp/overcooked_env.py
mesutyang97: I see. I can update it. But I do want to get a confirmation from Nathan that this will not affect his experiments (which I believe should be the case)

micahcarroll (approved):
Seems like a relatively minor PR. LGTM on a high level.

PR #69, opened 2020-11-26
Layout Evaluator Changes

jocelyn-chen:
These changes update the terrain_analysis function in layout_evaluator.py to properly take in a terrain matrix and generate possible action paths for two agents to make a soup.  The terrain_analysis function generates various action paths to make and deliver a 3-onion soup given a layout using various functions and infrastructure in layout_evaluator.py. The returned action paths can then be passed into an entropy function to evaluate the difficulty of a path/layout.

Key changes:

- addition of path_to_actions method that takes in two location paths and a terrain matrix and generates the action paths corresponding to the location paths. Undefined locations are used to determine when agents are performing the 'interact' action, otherwise are labeled as undefined actions.
- addition of path_to_actions_with_padding method that takes in two location paths, a terrain matrix, and two previous locations (one for each agent). This method is similar to the path_to_actions method above with the difference being this method ensures no undefined actions in the action list. This method does this by replacing undefined actions with padding that either pads with stay actions, or in the case in which an agent has a desired end point to pick up an item from a counter, it will pad with actions that will move the agent to its desired position. 
- the OvercookedMLASearchNode has been modified to store action lists and a single previous location for each agent instead of storing location lists as well as converting location lists to action lists to add them on in the update_from_search_node function
- in the perform_mla method, modifications have been made for collision detection and avoidance. Collision avoidance has been set to move the nonprimary agent to an empty neighboring spot if a collision is detected and to not add the mla path if a collision is unavoidable.
- the terrain_analysis function now takes in a terrain_mtx, and runs through the search process to generate possible action paths for the two agents to (possibly) collaboratively make a soup with three onions. Their action paths are returned in a dictionary.
- addition of a calculate_entropy_from_path which takes in an action list and a parameter ro and calculates the entropy of the given path 
- tests have been added to test all the functionality and changes listed above in the layout_evaluator_test.py file

### DIFF
--- a/src/overcooked_ai_py/mdp/layout_evaluator.py
+++ b/src/overcooked_ai_py/mdp/layout_evaluator.py
@@ -117,7 +117,30 @@ def path_to_actions(path_0, path_1, terrain_mtx):
     assert len(actions_0) == len(actions_1), " resulting actions should have the same length. Please pad if otherwise"
     return actions_0, actions_1
 
+def connect_action_path(action_paths):
+    """
+        Arguments:
+            action_paths (list of lists): a list of action paths to connect together
+        return:
+            connected action paths
+            currently just concatenates paths regardless of undefined actions
+        TODO: fill in undefined actions
+    """
+    total_path = []
+    for path in action_paths:
+        total_path += path
+    return total_path
+
 def calculate_entropy_of_path(path, ro):
+    """
+        Arguments:
+            path (list): a list of tuble for locations of agent 0
+            ro: parameter used in calculation of entropy
+        return:
+            the calculated entropy of the path
+            entropy of a sequence of one action 'a' repeated 'n' times is -ln(n) + ln(ro)
+            total entropy is the sum of all the sequences in the path
+    """
     total_entropy = 0
     const_entropy = np.log(ro)
     curr_length = 1
@@ -129,6 +152,7 @@ def calculate_entropy_of_path(path, ro):
         if action == curr_action:
             curr_length += 1
         else:
+            # if the sequence of actions that is ending is undefined, then we don't add the entropy
             if curr_action == UNDEFIND_ACTION:
                 curr_length = 1
                 curr_action = action
@@ -138,7 +162,7 @@ def calculate_entropy_of_path(path, ro):
             curr_length = 1
             curr_action = action
 
-    # add entropy for last sequence of actions
+    # add entropy for last sequence of actions if the last sequence is undefined
     if curr_action != UNDEFIND_ACTION:
         entropy = -np.log(curr_length) + const_entropy
         total_entropy += entropy
@@ -747,9 +771,6 @@ def terrain_analysis(terrain_mtx, silent = True):
 
     stage_score = []
 
-    player_1_action_path = []
-    player_2_action_path = []
-
     p0_i, p0_j = p0_starting
     p1_i, p1_j = p1_starting
     terrain_mtx_rep = copy.deepcopy(terrain_mtx)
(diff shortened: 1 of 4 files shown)


mesutyang97 (approved):
Looks good to me! Thanks for all the hard work!

src/overcooked_ai_py/mdp/layout_evaluator.py
```diff
@@ -564,35 +909,113 @@ def perform_mla(feature_locations, prev_mla_dict, walk_graph, handover_graph, te
...
-        # Return all (i, j) locations of feature
-        l, w = len(terrain_mtx), len(terrain_mtx[0])
-        res = []
-        for i in range(l):
-            for j in range(w):
-                if terrain_mtx[i][j] == feature:
-                    res.append((i, j))
-        return res
 
+def terrain_analysis(terrain_mtx, silent=True, best_only=True):
```
micahcarroll, line 1011: This function is about 400 lines long – in software engineering it is usually recommended to keep functions to 10 or 20 lines at most, and to break down individual tasks into other functions if necessary (this usually enables to reduce repeats significantly). For research engineering, I think this advice still applies, although a bit more lax. I try to keep my functions under 50 lines unless there are what I feel like very good reasons to keep it slightly longer than that. The main issue is that long functions tend to be much harder to follow and debug.

Also, the current docstring isn't informative about what is done by the function: my understanding is that you are trying to determine various (reasonable) ways of creating and delivering 1 3-onion soup, and returning the possible plans and their lengths?

src/overcooked_ai_py/mdp/layout_evaluator.py
```diff
@@ -30,14 +86,208 @@ def path_to_actions(path_0, path_1, terrain_mtx):
...
+    total_path = []
+    for path in action_paths:
+        total_path += path
+    return total_path
+
+def calculate_entropy_of_path(path, ro):
+    """
+        Arguments:
+            path (list): a list of tuble for locations of agent 0
+            ro: parameter used in calculation of entropy
```
micahcarroll, line 255: I'm a bit confused as to why `ro` is being used here (usually spelt rho btw 😛). My understanding is that in this function, 𝜌 is only used to add a constant to the entropy calculation. Can't we just remove it?

micahcarroll (approved):
Hi all, sorry for the late review on this! First of all, thanks Jocelyn for your work!

Also, I have never actually seen this file before (because it was on a unreviewed branch by Mesut), so I have less context then usual to evaluate what is being done. Would it be possible to edit the PR summary to have a higher-level summary of what the entire file is supposed to do?

I'll have a stab at it here, but let me know if I'm missing something: this entire file, `layout_evaluator`, has the purpose of generating statistics (?) for layouts (might be good to rename, as that seems different from the original intent of evaluating the difficulty of layouts). Most of the infrastructure in the file seems mainly aimed at bolstering the `terrain_analysis` function, which generates possible plans to create and deliver a 3-onion soup in the layout at hand. This might be then further used to calculate the entropy of the path.

I've added a couple of relatively high level comments in the code! As this is going to be merged in Mesut's branch (rather than a main branch), I think it's fine for the PR to not be spotless!

src/overcooked_ai_py/mdp/layout_evaluator.py
jocelyn-chen: Thanks for letting me know its rho not ro :). Rho is a constant that is added to every sequence of consecutive identical actions. So for a path that consists purely of moving up, the rho entropy would only be added once. However for a path that consists of moving up and right with several turns, the rho entropy would be added several times. It is essentially a way to add in information about the number of turns in a given path.

src/overcooked_ai_py/mdp/layout_evaluator.py
jocelyn-chen: Yup, your understanding of the function is correct. For now, I've added more information to the docstring about what is being returned by the function. After the merge, I can put in some work to refactor and break down the method.

src/overcooked_ai_py/mdp/layout_evaluator.py
micahcarroll: Do you know where you/Mesut got this formulation of path entropy from? I don't think I've seen it before

jocelyn-chen: I'm not sure where Mesut got it. I believe he may have formulated it himself.

PR #70, opened 2021-01-17
Waypoint Extraction

jocelyn-chen:
Edited the layout_evaluator.py to construct and include information about waypoints for paths.

A waypoint is considered to be when an event has happened such as when an onion is picked up or potted, or a soup starts cooking or is delivered. A waypoint entry is of the format [event_name, primary_agent_idx, agent_loc, item_loc]. 
- event_name: the name of the event (ex: 'onion_pickup', 'soup_delivery')
- primary_agent_idx: the index of the agent that executed the event
- agent_loc: the location of the agent when they executed the event
- item_loc: the location of the item needed to execute the event (ex: onion dispenser, pot, dish dispenser)
Both the agent location and item location are needed to determine the event occurrence as neither one is uniquely identifying by itself. 

Changes made:
- OvercookedMLASearchNode: added waypoints field to keep track of waypoints
- perform_mla function: added construction of waypoints in this function
- terrain_analysis function: now returns a list of waypoints for each possible path
- parse_event_name function: given an event_name, parses it to remove unnecessary information and maps it to match the event names that will be used in trajectories generated from human data

### DIFF
--- a/src/overcooked_ai_py/mdp/layout_evaluator.py
+++ b/src/overcooked_ai_py/mdp/layout_evaluator.py
@@ -419,7 +419,7 @@ class OvercookedSearchNode:
 
 class OvercookedMLASearchNode:
     def __init__(self, primary_idx, agent_0_loc, agent_1_loc, pot_loc, agent_0_act_dict, agent_1_act_dict,
-                 num_counter_ops, terrain_mtx):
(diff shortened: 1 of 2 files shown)


mesutyang97 (approved):
Thanks! Merging!

PR #72, opened 2021-03-28
terrain_analysis bug fix

jocelyn-chen:
Fixes a bug in which the function terrain_analysis in layout_evaluator.py was modifying the terrain_mtx parameter which led to inconsistent behavior when calling terrain_analysis several times.


mesutyang97 (approved):
Thanks for catching the issue!

PR #76, opened 2021-05-25
Tomato featurization

nathan-miller23:
Updates in `featurize_state` to account for both tomatoes and variable length pots

* `featurize_state` now takes a `num_pots` parameter. The state of the closest `num_pots` pots is encoded into higher dimensional features
  * Before, closest pot of every "type" (empty, 1 onion, 2 onion, full) was encoded
  * Now, there are exponentially more "states" a pot can be in, so fully encode the state of the closest  `num_pots` pots intead
 * Support for tomatoes!
   * Symmetric encodings to onions are added for tomatoes (closest, closest dispenser, etc)
 * Added corresponding tests to verify encoding works across plethora of tomato and non-tomato layouts


nathan-miller23:
* Fixed bug where calls to `OvercookedEnv.featurize_state_mdp` were passing in `self.horizon` as the `num_pots` param, resulting in massive featurized vector sizes
* Added `overcooked_ai_py/mdp/overcooked_trajectory.py` file
  * Codified the 'overcooked trajectory' schema
  * Only has one util function so far
  * Primitive precursor to full-fledged object-oriented trajectory handling (this has been a want-to-have on the backlog for almost a year)

src/overcooked_ai_py/mdp/overcooked_env.py
```diff
@@ -4,6 +4,7 @@
 from overcooked_ai_py.utils import mean_and_std_err, append_dictionaries
 from overcooked_ai_py.mdp.actions import Action
 from overcooked_ai_py.mdp.overcooked_mdp import OvercookedGridworld, EVENT_TYPES
+from overcooked_ai_py.mdp.overcooked_trajectory import TIMESTEP_TRAJ_KEYS, EPISODE_TRAJ_KEYS, DEFAULT_TRAJ_KEYS
```
mesutyang97, line 7: Thanks for this refactoring here!

src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -2025,32 +2083,43 @@ def make_closest_feature(idx, name, locations):
 
                 all_features["p{}_wall_{}".format(i, direction)] = [0] if feat == ' ' else [1]
 
+        # Convert feature dict created above into np array for each player
+        # NOTE: This is the portion that assumes n=2 players
         features_np = {k: np.array(v) for k, v in all_features.items()}
-
+        features_np = {k : np.expand_dims(v, 0) if not v.shape else v for k, v in features_np.items()}
```
mesutyang97, line 2089: Could you please explain a bit what's actually going on here?

mesutyang97 (commented):
Thanks @nathan-miller23 for the great work! Overall looks really good! Just hope to get clarification on one minor thing as commented.

src/overcooked_ai_py/mdp/overcooked_mdp.py
nathan-miller23: This was there to convert 'scalar' features into single dimension numpy arrays. For example, the feature `p0_closest_soup_num_onions` might be set to the value of 3 in the feature dictionary. Calling `np.array(3)` returns a zero dimension array, so the `expand_dims` was necessary to get it into a shape compatible with all the other 1D array features.

However, I changed the code to avoid this confusion by setting `p0_closest_soup_num_onions = [3]` (in the above example) from the beginning

nathan-miller23:
* Fixed bug in `featurize_state` that caused pots to be incorrectly encoded when `num_pots != 2`
* Added more robust tests for `featurize_state` that iterate over `num_pots`
* Deprecated `mdp.featurize_state_shape` (and `mdp.lossless_state_encoding`) property
  * New use is `mdp.get_featurize_state_shape` (and `mdp.get_lossless_state_encoding_shape`)
  * This is so `num_pots` can be passed into the featurize_shape method, as `num_pots` impacts encoding shape
  * Old properties still supported (for now) they just raise a deprecation warning

src/overcooked_ai_py/mdp/overcooked_mdp.py
nathan-miller23: I also caught a bug in this function, so now the code here is a little outdated

src/overcooked_ai_py/mdp/overcooked_env.py
nathan-miller23: Still think we should aim for a full-fledged `Trajectory` class at some point, but this should serve as a nice middle ground for now!

src/overcooked_ai_py/data/testing/test_display_phi/exected.txt
```diff
@@ -0,0 +1,309 @@
+X       X       P       X       X
```
micahcarroll, line 1: The name of this file seems like a typo? :)

nathan-miller23: fixed :P

src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1953,30 +1959,119 @@ def process_for_player(primary_agent_idx):
...
+        num_pot_features = 10
+        base_features = 28
+        total_features = self.num_players * (num_pots * num_pot_features + base_features)
+        return (total_features,)
+
+    def featurize_state(self, overcooked_state, mlam, num_pots=2, **kwargs):
         """
-        Encode state with some manually designed features.
-        NOTE: currently works for just two players.
+        Encode state with some manually designed features. Works for arbitrary number of players
```
micahcarroll, line 1980: Maybe add a comment explaining what num_pots does?

src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1989,68 +2084,82 @@ def make_closest_feature(idx, name, locations):
...
+            all_features = concat_dicts(all_features, make_closest_feature(i, player, "dish", self.get_dish_dispenser_locations() + counter_objects["dish"]))
+            all_features = concat_dicts(all_features, make_closest_feature(i, player, "soup", counter_objects["soup"]))
+            all_features = concat_dicts(all_features, make_closest_feature(i, player, "serving", self.get_serving_locations()))
+            all_features = concat_dicts(all_features, make_closest_feature(i, player, "empty_counter", self.get_empty_counter_locations(overcooked_state)))
+
+            # Closest pots info
+            pot_locations = self.get_pot_locations().copy()
+            for pot_idx in range(num_pots):
+                _, closest_pot_loc = mlam.motion_planner.min_cost_to_feature(player.pos_and_or, pot_locations, with_argmin=True)
+                pot_features = make_pot_feature(i, player, pot_idx, closest_pot_loc, pot_states)
```
micahcarroll, line 2099: Just checking that this currently supports the case in which there might be a variable number of pots across different layouts you are interested in, and you want the featurization shape to be the same across those different cases.

For example, if you have two layouts (one with 1 pot and one with 2) and you want to do BC on both layouts simultaneously (rather than training 2 separate agents), you would probably want to set num_pots=2 and then have the 2nd pot slot in the featurization of the first layout have some null token (all 0s or something).

min_cost_to_feature will not return a pot_location, and make_pot_feature will return all zeros?

micahcarroll (approved):
LGTM! 🙌 Thanks Nathan!

src/overcooked_ai_py/mdp/overcooked_mdp.py
nathan-miller23: Correct! In that case, `closest_pot_loc` will be `None`, so `make_pot_feature` will put in all zeros for the second pot. Building off of your comment above, I think it's a good idea to add a thorough description of the feature encodings, what exactly `num_pots` does, and how it handles the 'there is no pot to encode' case at the top of the function in a docstring would be worthwhile

mesutyang97 (approved):
Overall looks great! Thanks Nathan

src/overcooked_ai_py/mdp/overcooked_mdp.py
nathan-miller23: Added an in depth docstring explaining what each param does and what features exactly are encoded

PR #80, opened 2021-11-13
Correct citation for Fontaine 2021

btjanaka:
Hi @micahcarroll, I noticed my name was misspelled in the list of citations (I think this happened because the original arXiv citation misspelled my name). Please accept this PR which corrects the spelling. Thanks!


micahcarroll:
Thank you for the PR, and sorry about this – you're quite right in that I believe the problem came from the name spelling from the arxiv version!

PR #102, opened 2022-11-15
Wandb + test optimization

jyan1999:
# Description

Added Wandb support. Metrics are logged to wandb if verbose = True. 
Fixed test_ppo_bc bug and minor optimizations
Optimized testing thresholds for PPO tests.

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

# Checklist:


src/human_aware_rl/ppo/ppo_rllib_from_params_client.py
```diff
@@ -391,7 +391,13 @@ def _env_creater(env_config):
 
 
 def run(params):
+    layout = "MyPPOModel"
```
micahcarroll, line 394: Is this the right variable name? Doesn't seem like a layout?

src/human_aware_rl/ppo/ppo_rllib_client.py
```diff
@@ -341,11 +341,17 @@ def my_config():
 
 
 def run(params):
+    layout = params["environment_params"]["mdp_params"]["layout_name"]
+    conf = params["experiment_name"]
+    if params["verbose"]:
+        import wandb
+
+        wandb.init(project=layout, sync_tensorboard=True)
```
micahcarroll, line 349: Wandb project should be the same across all runs probably – this can be a hardcoded global variable imported from utils?

src/human_aware_rl/ppo/ppo_rllib_test.py
```diff
@@ -251,9 +250,7 @@ def test_ppo_fp_sp_no_phi(self):
             options={"--loglevel": "ERROR"},
         ).result
         # Sanity check (make sure it begins to learn to receive dense reward)
-        self.assertGreaterEqual(
-            results["average_total_reward"], self.min_performance
```
micahcarroll, line 255: Ise `self.min_performance` still assigned to a value somewhere? What value does it have? It might be good to clean that up

src/human_aware_rl/ppo/ppo_rllib_test.py
```diff
@@ -320,14 +316,13 @@ def test_ppo_bc(self):
...
             "evaluation_interval": 5,
             "verbose": False,
+            "layout_name": "asymmetric_advantages_tomato",
         }
         results = ex.run(
             config_updates=config_updates, options={"--loglevel": "ERROR"}
         ).result
         # Sanity check
-        self.assertGreaterEqual(
-            results["average_total_reward"], self.min_performance
```
micahcarroll, line 329: Same here :)

micahcarroll (changes requested):
Overall LGTM with some small changes!

src/human_aware_rl/ppo/ppo_rllib_from_params_client.py
jyan1999: Ah yeah I was going to bring this up during the meeting today. So it looks like ex_fp (the experiment from ppo_rllib_from_params_client.py) supports creating a MDP not from a specific layout but from a "mdp_params_schedule_fn". So I just tried to get a placeholder for the layout name, so it is grouped correctly in wandb. It is not actually a layout and I should have probably picked a better name.

src/human_aware_rl/ppo/ppo_rllib_client.py
jyan1999: Wandb will overlay graphs from the same project on top of each other, so I thought it is probably a good idea to log the experiments based on layouts?

src/human_aware_rl/ppo/ppo_rllib_test.py
jyan1999: I changed the performance thresholds for those with relatively large performance differences. Others have pretty small differences so I just kept the old performance threshold = 5 as sanity checks. Self.min_performance is set in the __init__ method.

src/human_aware_rl/ppo/ppo_rllib_client.py
micahcarroll: I think that's fine – it's very easy to filter out the graphs that one doesn't want to look at or group them as one wishes! We can look through an example together at some point! 😄

src/human_aware_rl/ppo/ppo_rllib_test.py
micahcarroll: I see, makes sense! Maybe worth adding this as a comment in the code just have it clear why these are hardcoded to be other values

src/human_aware_rl/ppo/ppo_rllib_from_params_client.py
micahcarroll: Probably this is not relevant after we've dealt with the case below 😄

src/human_aware_rl/ppo/ppo_rllib_client.py
jyan1999: Got it. I'll remove those.

micahcarroll (approved):
LGTM! Thanks Jackson!

PR #103, opened 2022-11-19
Added old_dynamics flag

### LINKED ISSUE #101: Press the space bar to start cooking OR start directly with three onions on the pot
Dear developer,
Which line of overcooked Env to alter the config that press the space bar to start cooking OR start directly with three onions on the pot?
Thanks

jyan1999:
# Description

Added the old_dynamics flag to enable cooking without interaction. 

Fixes #101 

## Type of change

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works -- qualitatively checked performance differences on layouts from 2019

### DIFF
--- a/src/overcooked_ai_py/mdp/overcooked_mdp.py
+++ b/src/overcooked_ai_py/mdp/overcooked_mdp.py
@@ -1390,6 +1390,7 @@ class OvercookedGridworld(object):
         assert not self.is_terminal(
             state
         ), "Trying to find successor of a terminal state: {}".format(state)
+
         for action, action_set in zip(joint_action, self.get_actions(state)):
             if action not in action_set:
                 raise ValueError(
@@ -1397,13 +1398,11 @@ class OvercookedGridworld(object):
                 )
 
         new_state = state.deepcopy()
-
         # Resolve interacts first
         (
             sparse_reward_by_agent,
             shaped_reward_by_agent,
         ) = self.resolve_interacts(new_state, joint_action, events_infos)
-
         assert new_state.player_positions == state.player_positions
         assert new_state.player_orientations == state.player_orientations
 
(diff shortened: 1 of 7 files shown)


src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1136,6 +1144,8 @@ def __init__(
         self._opt_recipe_discount_cache = {}
         self._opt_recipe_cache = {}
         self._prev_potential_params = {}
+        # determines whether to start cooking automatically
```
micahcarroll, line 1147:
```suggestion
        # determines whether to start cooking automatically once 3 items are in the pot
```

src/overcooked_ai_py/mdp/overcooked_mdp.py
```diff
@@ -1506,8 +1516,11 @@ def resolve_interacts(self, new_state, joint_action, events_infos):
                 player.set_object(obj)
 
             elif terrain_type == "P" and not player.has_object():
-                # Cooking soup
-                if self.soup_to_be_cooked_at_location(new_state, i_pos):
+                # Cooking soup only if we are using the new dynamics
```
micahcarroll, line 1519:
```suggestion
                # An interact action will only start cooking the soup if we are using the new dynamics
```

micahcarroll (approved):
This looks great! Thanks so much Jackson!!

micahcarroll:
Actually, can we add a simple MDP test that makes sure that both old_dynamics=True and False behave as intended?

jyan1999:
> When you say that "qualitatively checked performance differences on layouts from 2019", what do you mean?

Oh yeah so while I was working on fixing BC agent training issues I realized most training done on the 2019 layouts don't get any rewards since the cooking doesn't start. The agents put all the ingredients in the pot and just let them sit there. Now I know what is causing that behavior I checked whether the performance has improved if I changed the dynamics, which it has. I checked the visualization, and it looks like we are getting the desired behaviors. 

> trying to start the soup cooking from a state that just has 1 item in the pot with an interact – and it working in one case and not the other. And on the flipside, putting a third item in a pot and seeing that it's already started cooking with old dynamics and hasn't in the new dynamics

I'll add these. I was trying to figure out a way to test this feature and this sounds like a really good way to do it.

micahcarroll:
Looks great!! Thanks @jyan1999!!

PR #116, opened 2023-03-09
Reproduce agents

jyan1999:
# Description

Finished up setting up new models and put the experiments results in a separate google drive folder

Fixes # (issue)

## Type of change

- [x] This change requires a documentation update

# Checklist:


README.md
```diff
@@ -161,6 +161,7 @@ Overcooked_demo can also start an interactive game in the browser for visualizat
 
 The raw data used in training is >100 MB, which makes it inconvenient to distribute via git. The code uses pickled dataframes for training and testing, but in case one needs to original data it can be found [here](https://drive.google.com/drive/folders/1aGV8eqWeOG5BMFdUcVoP2NHU_GFPqi57?usp=share_link) 
 
+The existing agents that are available to play with in the overcooked_demo module are taken from [this group](https://drive.google.com/drive/folders/1CI3rLZg5wpjpoTDBi1M1qn8xvjqhk5Md?usp=share_link) of reproduced agents. To reproduce the results locally, you can run [this script](https://github.com/HumanCompatibleAI/overcooked_ai/blob/master/src/human_aware_rl/ppo/run_experiments.sh).
```
micahcarroll, line 164: I would move this to the overcooked-demo repo README!

src/overcooked_demo/README.md
```diff
@@ -58,12 +58,37 @@ If a more complex or custom loading routing is necessary, one can subclass the `
 
 ## Use the human vs. human game mode.
 
-With the Overcooked demo, you can test the interaction between two human players. To do this, you need to deploy this code on the server (https://docs.docker.com/language/python/deploy/). 
+### 1): Using Docker Compose CLI to deploy on EC2
+With the Overcooked demo, you can test the interaction between two human players. To do this, you need to deploy this code on the server (hsttps://docs.docker.com/language/python/deploy/).
```
micahcarroll, line 62:
```suggestion
With the Overcooked demo, you can test the interaction between two human players. To do this, you need to deploy this code on a server (hsttps://docs.docker.com/language/python/deploy/). 
```

"a" server? what do you mean by "the"?

src/overcooked_demo/README.md
```diff
@@ -58,12 +58,37 @@ If a more complex or custom loading routing is necessary, one can subclass the `
...
+With the Overcooked demo, you can test the interaction between two human players. To do this, you need to deploy this code on the server (hsttps://docs.docker.com/language/python/deploy/). 
 After successful deployment, the first user should open http://[server_ip_address]/, select the human keyboard input for both players and click on "Create game". If everything has been successful, he will receive a message: "Waiting for game to start".
 Another user should open a page at http://[server_ip_address]/psiturk to start the game.  
 
 If you want to run a test on a local computer, you should use "localhost" instead of "server_ip_address" and open the corresponding links in different tabs.
 
+### 2): Setup the repo on a cloud server 
+One also has the option to setup the repo again on a cloud server. This allows more flexible customization and easier spin-up. The following steps demonstrate how one can steup the repo on an AWS EC2 instance 
+
+1: Get an [AWS account](https://aws.amazon.com/free/)
```
micahcarroll, line 71: Does it work with the free account? If not, let's not link to that. Also worth mentioning that it doesn't require using AWS but we only talk about AWS instructions

src/overcooked_demo/server/game.py
```diff
@@ -37,18 +37,18 @@ def fix_bc_path(path):
     For now the solution is to include the saved BC model and fix the relative path to the model in the config.pkl file
     """
 
-    import pickle
+    import dill
```
micahcarroll, line 40: Why did you switch to dill? Just curious!

micahcarroll (commented):
Left some comments! Thanks @jyan1999!! 🚀

src/overcooked_demo/server/game.py
jyan1999: Okay so this is a long story.... I was running into a weird bug that no agents would load when running the web demo from the newly reproduced agents, and I tried a lot of things, including changing pickle to dill for the reason [documented here](https://github.com/HumanCompatibleAI/overcooked_ai/blob/15e50b7228f2b58220636e411905a668d79d12d5/src/human_aware_rl/rllib/rllib.py#L822). 
It turns out that the actual reason was network delay, by which I mean since I was downloading a lot of checkpoints files from the server at once, and they didn't finish downloading until a couple of minutes later, but the directories and config.pkl files became visible before the downloads were complete, meaning the reason why the load failed was because the files were missing contents......
Now everything works and I just didn't bother changing it back to pickle since I found this comment and it seemed reasonable.

src/overcooked_demo/README.md
jyan1999: Yeah it should work fine, and the signup portal is the same for all types of accounts. With a free account you also get access to not free instances (it is just you can't use the really good ones like P-type instances with GPUs without building a history of good credits), but for the purpose of running overcooked games a free account should be sufficient.

src/overcooked_demo/README.md
jyan1999: hmmm I didn't write this part. I think it was there on the old readme and I just copied and pasted it. I guess they are synonymous in this context but ..... I'll just change it.

src/overcooked_demo/README.md
```diff
@@ -68,7 +72,7 @@ If you want to run a test on a local computer, you should use "localhost" instea
 ### 2): Setup the repo on a cloud server 
 One also has the option to setup the repo again on a cloud server. This allows more flexible customization and easier spin-up. The following steps demonstrate how one can steup the repo on an AWS EC2 instance 
 
-1: Get an [AWS account](https://aws.amazon.com/free/)
+1: Get an [AWS account](https://aws.amazon.com/free/). Note any cloud server should work and it is by no means limited to AWS, and the instruction provided here is only meant to give an example for how it can be done.
```
micahcarroll, line 75:
```suggestion
1: Get an [AWS account](https://aws.amazon.com/free/). Note that any cloud provider should work – the instructions provided here for AWS are only meant as an example.
```

src/overcooked_demo/server/game.py
micahcarroll: Damn that sounds very annoying! 😅

micahcarroll (approved):
LGTM

micahcarroll:
I'll merge this for now, and we can update the PPO_BC agents next in a separate PR!

PR #119, opened 2023-03-23
Reproduce PPO_BC agents

jyan1999:
# Description

Reproduced PPO-BC agents with comparable performances to those in the paper. Also fixed miscellaneous issues:
1): Changed default fps to 6
2): Included missing html and js files from previous commit

Fixes # (issue)

## Type of change

# Checklist:


micahcarroll:
Include bash script for reproduce agents

src/human_aware_rl/ppo/run_ppo_bc_experiments.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# This file contains the script to generate the ppo self-play agents for the 5 classic layouts
+
+#First we need to create the BC-agents, the agent will be saved in the bc_runs directory under human_aware_rl/bc_runs directory
+python "$(dirname "$0")/../imitation/reproduce_bc.py"
```
micahcarroll, line 5: Maybe specify which directory this bash script should be run from as a comment in the file too?

src/overcooked_demo/server/static/js/predefined.js
micahcarroll, line 1: What's this file for? Maybe add a comment at the top

src/overcooked_demo/server/static/templates/predefined.html
micahcarroll, line 1: Same here

micahcarroll (approved):
Except for very minor comments, this all looks great! Thanks so much @jyan1999 !!

src/human_aware_rl/ppo/run_ppo_bc_experiments.sh
jyan1999: Got it. It actually doesn't really matter where the bash script is run since this path is relative to where this script is, not where it is being called. I'll add some comments clarifying that.

micahcarroll: Ah makes sense – my bad for not realizing that

PR #121, opened 2023-04-09
Graph

jyan1999:
# Description

Added performance graph to the human_aware_rl readme and codes that produce it

Also removes the stale issue checker, since I feel it might be better to let the issue and pr be visible as a reminder and manually close them when needed.

Fixes # (issue)

## Type of change

# Checklist:


.github/workflows/stale-issues.yml
micahcarroll, line 1: Changed your mind? LOL

src/human_aware_rl/README.md
```diff
@@ -75,6 +75,18 @@ You can also try to train agents for 5 classical layouts
...
 
+## Experiments results
+
+This graph shows the performance of reproduced agents in a similar style as original ones in the Neurips 2019 paper
+
+![Performance graph](./graph.jpg)
+
+It was created with the reproduced agents from running run_experiments.sh and run_ppo_bc_experiments.sh
+
+The evaluation and plotting code can be found under human_aware_rl/ppo directory
```
micahcarroll, line 86:
```suggestion
This graph shows the performance of reproduced agents in a similar style as the original ones in the Neurips 2019 paper.

![Performance graph](./graph.jpg)

It was created with the reproduced agents from running `run_experiments.sh` and `run_ppo_bc_experiments.sh`.

The evaluation and plotting code can be found under the `human_aware_rl/ppo` directory.
```

src/human_aware_rl/ppo/plot_graph.py
```diff
@@ -0,0 +1,186 @@
...
+            offset,
+            results_0[i],
+            width,
+            color=color,
+            edgecolor="gray",
+            lw=1.0,
+            zorder=0,
+            label=algo,
+            linestyle=":",
+            yerr=stds[i],
```
micahcarroll, line 112: If you delete these last commas, does Black formatting let it fit on one line?

src/human_aware_rl/README.md
```diff
@@ -75,6 +75,18 @@ You can also try to train agents for 5 classical layouts
 Please choose the `num_workers` parameters to be less than the number of CPUs on your machine. For more information
 check the ray [documentation](https://docs.ray.io/en/latest/rllib/rllib-training.html)
 
+## Experiments results
+
+This graph shows the performance of reproduced agents in a similar style as original ones in the Neurips 2019 paper
+
+![Performance graph](./graph.jpg)
```
micahcarroll, line 82: Graph is missing :)

micahcarroll (changes requested):
LGTM, but missing graph! Thanks so much @jyan1999!!

src/human_aware_rl/README.md
jyan1999: oops, okay it is added now

src/human_aware_rl/ppo/plot_graph.py
jyan1999: hmmm no it still formats it like this. I'll see if I can get around this without having to change Black to reformat every files

micahcarroll: Don't worry, not worth looking for a workaround! Black is an "uncompromising formatter" for a reason :)

src/human_aware_rl/graph.jpg
micahcarroll, line 1: One change we probably want to do is have the red dotted lines be above just the relevant orange bar (as in the neurips graphs).

On another note, it does seem like some of the orange bars are a bit lower than the ones in the original paper 😅 that's ok, but a lil sad

micahcarroll (approved):
Very minor changes to red dotted lines still to be added, after which can be merged!

.github/workflows/stale-issues.yml
jyan1999: Yeah I thought this was going to be cool since I see other repos have it, but on second thoughts given the numbers of issues and PR's I don't think this is very necessary...... 
(Plus I didn't like how each issue now has a ddl now and I have to deal with them before the github bot fails me...)

src/human_aware_rl/graph.jpg
jyan1999: Oh okay. I was comparing it to the graph in the notebook and didn't realize it was different from the ones in the paper. I'll fix that.

About the performance difference, I also noticed it after making the graph, and I think it might have something to do with how the BC/human proxy models were trained. I didn't do a sweep when training the BC and human proxy agents and just used the default parameters, so the agents themselves might not have been optimized. The BC+PPO_BC performed well enough, so I thought the training parameters were good, but there might be some amount of overfitting so there is a drop when pairing with the human proxy models. 

It is also the case that the standard deviation/errors for ppo_bc + hp runs are bigger than before (which could also hint towards overfitting?), so the variability in the output might also dragged the mean performance down a bit.

.github/workflows/stale-issues.yml
micahcarroll: Looool yeah makes sense

micahcarroll:
LGTM, thanks @jyan1999 !!

PR #141, opened 2024-04-15
Fixed typo in p1 feature array

lewj85:
# Description

Fixed a typo in the absolute position of the partner player (p1) for their feature array.


micahcarroll (approved):
Thanks so much for catching this! 😬 😅

micahcarroll:
As long as you don't mind not getting attribution for this, given that the conflicts are massive (due to the fact that your making a change from a fork which is 500+ commits behind), I'm thinking of directly modifying the code myself (I can leave a comment thanking you! :) )

PR #158, opened 2025-02-25
WIP migrate to uv package manager, fixing package compatibility issues

joshcarp:
# Switching to uv, overcooked-ai version 2.0.0

Should make package versioning much more stable (hopefully). 

Some things have broken since last time I've looked at the repo. The main one is that GPU training doesn't seem to be working on my server (I think our server's GPU NVIDIA drivers are too new for the tensorflow version used in this repo?). It's annoying because it means that it's harder to run the replication scripts and make sure things still work.

That said, after this PR, all tests pass.

(This was written by @micahcarroll, took over @joshcarp's PR)


micahcarroll:
Hey @joshcarp, if you do end up getting things to work with uv, I would be very happy to merge this! Just wanted to let you know ahead of time!

